PR #20, opened 2015-03-05
Added trailing slashes to urls

kavdev:
It's common convention and a Django recommendation to have urls with trailing slashes.


haxoza:
Thanks for this pull request. You're totally right.

The reason why there are no trailing slashes now is that we created this code as a backend for AngularJS application which didn't support trailing slashes on that time. However it does support from version 1.3., so it shouldn't be a problem.

My only concern is backward compatibility. If merge this pull request we should bump the version up to 0.3.0.

@konradhalas what would you say?

konradhalas:
@haxoza we should also update README

Issue #30, opened 2015-04-05
Multiple tokens per user so one can log in and log out from different devices independently


felipemoran:
It would be nice if every time a user logs in a new token was generated and sent. In this way a user could log in and out on his phone without getting logged out from his computer. But because of this a new method would be necessary, that would be being able to log out from all devices (making all tokens invalid)

konradhalas:
This is most important thing on our roadmap right now. Thank you for creating issue, I hope it will motivate us.

felipemoran:
Great, that would be a really nice feature. Thanks!

konradhalas:
According to decisions taken in #33 (summary: we will create separate app) I'm closing this issue.

Issue #28, opened 2015-03-16
Add support for BasicAuthentication backend


jxstanford:
I'm hesitant to call this an issue -- more of a curiosity.  Was there a reason that you chose not to implement the handling of authentication for the /login endpoint via the BasicAuthentication class in DRF?

haxoza:
We should be able to add support for BasicAuthentication quite easily. I hope we will implement and document it in the nearest future.

haxoza:
As long as #57 is fixed and closed the support for HTTP Basic Auth is fully working.

PR #33, opened 2015-04-17
Multi-client login - custom auth token

konradhalas:
This is PR for #30 issue. Pleas do **NOT** merge it now. This is just POC. We have to figure out a few things before merge.

Summary: with new custom authentication token user have to provide "client" ID when login, but when he log out - we will remove only token created for that client. He will be still "logged in" on other clients.

Remarks about new approach:
* there is new Django app with custom token model - `djoser.token` - I don't know is it the best name for it. Maybe `djoer.authtoken`? Or `djoser.multitoken`?
* I didn't add any new settings - you just have to install `djoser.token` and use new `djoser.token.authentication.TokenAuthentication` as your `DEFAULT_AUTHENTICATION_CLASSES`. It's possible due to `rest_framework.authentication.TokenAuthentication` API - it has `model` field. Please check magic in `djoser.utils.get_authentication_token_class` and `djoser.get_token_model`. I don't like it, but it works.
* please also review how I handle different login serializer for new token - `TokenAuthentication.get_login_serializer_class`
* probably `djoser.utils` is not the best place to put those magic methods but I don't have better idea now.
* maybe we should generate "client" ID when user didn't provide it?
* I think that we should remove `LOGIN_AFTER_ACTIVATION` / `LOGIN_AFTER_REGISTRATION` -  this approach doesn't fit well to current solution, and it doesn't make sense at all. 
* with new API we could add more important fields to our `Token` model - e.g. expiration date.
* we probably should also implement mechanism to limit number of clients (e.g `MAXIMUM_CLIENTS_NUMBER` with e.g. 10 as default).
* of course we have to prepare documentation for this PR, but I think we could do this at the end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,6 @@ DJOSER = {
     'SITE_NAME': 'Frontend',
     'PASSWORD_RESET_CONFIRM_URL': '#/password/reset/confirm/{uid}/{token}',
     'ACTIVATION_URL': '#/activate/{uid}/{token}',
-    'LOGIN_AFTER_ACTIVATION': True,
     'SEND_ACTIVATION_EMAIL': True,
 }
 ```
@@ -166,9 +165,6 @@ method and have [USERNAME_FIELD](https://docs.djangoproject.com/en/dev/topics/au
 and [REQUIRED_FIELDS](https://docs.djangoproject.com/en/dev/topics/auth/customizing/#django.contrib.auth.models.CustomUser.REQUIRED_FIELDS)
 fields.
 
-If `LOGIN_AFTER_ACTIVATION` is `True`, you will receive authentication token
-within response.
-
 #### `POST`
 
 URL: `/register/`
@@ -195,8 +191,6 @@ URL: `/register/`
         
         `{{ User.REQUIRED_FIELDS }}`
         
-        `auth_token` (if `LOGIN_AFTER_ACTIVATION` is `True`)
-
 ### Login
 
 Use this endpoint to obtain user [authentication token](http://www.django-rest-framework.org/api-guide/authentication#tokenauthentication).
@@ -253,10 +247,6 @@ URL: `/activate/`
 
     * status: `HTTP_200_OK` (success)
  
-    * data: 
-
-        `auth_token` (if `LOGIN_AFTER_ACTIVATION` is `True`)
-        
 ### Set username
 
 Use this endpoint to change user username (`USERNAME_FIELD`).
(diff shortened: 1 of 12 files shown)


konradhalas:
@rugala and @haxoza it will be awesome if you could leave your comments here when you find some spare time.

haxoza:
It took some time but here are my thoughts.

**Client field**
According to what we discussed the client field is not required because it's difficult to define what should be put in it. For example on iOS you cannot get any unique device ID. That means that anyway the client needs to generate some random sequence in order to fill client field or it should be randomly filled on server side.

We concluded that client field should be removed. However knowing that we client field was invented to protect against overflowing database with new tokens for the same users we decided to introduce expiration date to the token model. We should also provide a logic for removing expired tokens.

Setting like `MAXIMUM_CLIENTS_NUMBER` is also a good idea.

**The app name**
I would opt in for `multitoken` name. However I would move this django app to seprate project to not mess djoser core functionality.

When we add support for other auth backends (see issues) there is no reason to keep `multitoken` as a part of `djoser`. What's more I can imagine that we could provide more features to `multitoken` app like some storage or caching backends. This will be not connected with djoser directly and it can end up with mess in pull requests and issues on github.

**Authentication process**
In general it totally makes sense what you did with `TokenAuthentication` class in this new `token` app. However considering issues #28 and #40 I would suggest the new way of handling auth backends in djoser. Currently we have a lot of code depending on `TokenAuthentication`. Probably it's not good idea if we want to support other backends.

That's why I would remove all the code depending on `TokenAuthentication`, especially `UserLoginSerializer` and `LoginView` classes. With the new `token` app the responsibility how the authentication is handled should rely on this new app, not on `djoser`. It's handled exactly in the same way by other auth backends:

* [JWT auth](https://github.com/GetBlimp/django-rest-framework-jwt/blob/master/rest_framework_jwt/views.py#L40),
* [Token auth from DRF](https://github.com/tomchristie/django-rest-framework/blob/3.1.2/rest_framework/authtoken/views.py#L9),
* and what's funny `BasicAuthentication` backend does not require any obtain view.

**Other  stuff**
I agree we should remove `LOGIN_AFTER_ACTIVATION` / `LOGIN_AFTER_REGISTRATION` settings and support.

**Further steps**
Here is what I suggest to do now:

1. Extract new token app as a new project
2. Close this PR and branch.
3. Create new branch when we add support for other auth backend including new multitoken app.
4. When the all above works we will merge it to djoser and release new major version.

@konradhalas please deliberate :smile:

konradhalas:
Give me a month or so...

myquant:
any progress update?

haxoza:
@myquant not yet, but there will be really soon :smile: We've just released last djoser version before quite big refactoring which includes finishing this pull request.

haxoza:
I extracted `token` app created in this branch and pull request to the new project:

https://github.com/sunscrapers/django-rest-multitoken

PR #44, opened 2015-05-11
added signals for user activation and registration

barszczmm:


haxoza:
Maciej, thanks for submitting this feature! It would be great to have it documented someday.

PR #45, opened 2015-05-11
urlsafe_base64_decode() decodes to bytestring on Python 3 so force_text needs to be used

barszczmm:
See here: https://github.com/django/django/blob/stable/1.8.x/django/contrib/auth/views.py#L234


haxoza:
Thanks! We didn't know about that!

PR #50, opened 2015-06-02
change constants dynamic

ckdalsdhdp:
Hello
I would like to use the custom constants if failed to user login
but serializer class used fixed constants in validationError
so I changed constants to class variable
please give me feedback thanks :)

p.s.
sorry I can't write english well

### DIFF
--- a/djoser/serializers.py
+++ b/djoser/serializers.py
@@ -86,6 +86,8 @@ class UserRegistrationWithAuthTokenSerializer(UserRegistrationSerializer):
 
 
 class UserLoginSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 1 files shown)


ParkSori8:
very good

konradhalas:
Hi @ckdalsdhdp, thank you for your PR.

I like your idea, we should definitely provide easy way to override error messages in serializers. I only wonder is it the best solution. I'm trying to find how DRF handle such cases but I can't find anything in docs about it.

Nevertheless, we will merge your PR soon OR we will figure out sth similar (maybe we'll move those variables to settings?)

Issue #54, opened 2015-06-21
Remove support for DRF 2.4


haxoza:
See also #42.

haxoza:
@konradhalas please check this out and review :smiley: :coffee:

PR #56, opened 2015-06-24
#54 change wrong setting in Register section in readme

gitfred:
Changed `LOGIN_AFTER_ACTIVATION` to `LOGIN_AFTER_REGISTRATION` and remove some whitespaces in `README`

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,10 +12,10 @@ Instead of reusing Django code (e.g. `PasswordResetForm`), we reimplemented
 few things to fit better into [Single Page App](http://en.wikipedia.org/wiki/Single-page_application)
 architecture.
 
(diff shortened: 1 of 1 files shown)


haxoza:
Awesome! Thanks! :+1:

mcgeeco:
I implemented a login after activation myself, would that be a worthwhile feature?

konradhalas:
As I remember (@haxoza correct me if I'm wrong), we have plans to remove all `LOGIN_AFTER_*` settings, because 1) those settings increase code complexity too much, 2) you could just make login request after registration/activation. Does it make sense?

mcgeeco:
Yeah I suppose it makes sense to keep the complexity down, otherwise you could start implementing login after for everything.

konradhalas:
Indeed. Thanks.

PR #58, opened 2015-06-30
feat(djoser/):added PEP-8 formatting

aniketmaithani:


haxoza:
Thanks, but it's not so useful...

aniketmaithani:
@haxoza in my view the PEP8 formatting makes the code neat and readable in the first place. Not that it was not readable earlier but it's kind a like a benchmark.

PR #67, opened 2015-08-02
Allow email to be used for auth

mcastle:
This provides the option of using email instead of username for auth. It works nicely with auth packages such as django-allauth, which provides an email authentication backend.

### DIFF
--- a/djoser/serializers.py
+++ b/djoser/serializers.py
@@ -47,10 +47,11 @@ class LoginSerializer(serializers.Serializer):
     def __init__(self, *args, **kwargs):
         super(LoginSerializer, self).__init__(*args, **kwargs)
         self.user = None
-        self.fields[User.USERNAME_FIELD] = serializers.CharField(required=False)
+        self.fields[User.USERNAME_FIELD] = serializers.CharField(required=False, allow_blank=True)
+        self.fields['email'] = serializers.EmailField(required=False, allow_blank=True)
 
     def validate(self, attrs):
-        self.user = authenticate(username=attrs[User.USERNAME_FIELD], password=attrs['password'])
+        self.user = authenticate(**attrs)
         if self.user:
             if not self.user.is_active:
                 raise serializers.ValidationError(self.error_messages['inactive_account'])
--- a/testproject/auth_backends.py
+++ b/testproject/auth_backends.py
@@ -1,0 +1,19 @@
+from django.contrib.auth.backends import ModelBackend
+from django.contrib.auth import get_user_model
+
+
+class EmailAuthenticationBackend(ModelBackend):
+
+    def authenticate(self, **credentials):
+        ret = self._authenticate_by_email(**credentials)
+        return ret
+
+    def _authenticate_by_email(self, **credentials):
+        User = get_user_model()
+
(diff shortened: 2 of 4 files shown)


haxoza:
Thanks for pull request! However this change affects API of djoser so it would be good to have any tests for that. Especially when it comes to integration with third-party libraries like django-allauth.

Could you provide any tests for that feature and maybe some corresponding links to django-allauth documentation?

mcastle:
I've added some tests. The changes should work with any backend that allows email for auth, not just with django-allauth. And it doesn't prevent the default authentication backend from working (auth with a username and password).

Here's the link to the email auth backend for django-allauth (again, this pull is not specific to allauth, this is just an example):

https://github.com/pennersr/django-allauth/blob/master/allauth/account/auth_backends.py

haxoza:
Thanks for PR! I will get into it early next week!

mcastle:
No prob! Thanks for creating djoser!

konradhalas:
@haxoza where we are with this PR?

ashishmohite:
am i the only one who feels not just email or username but any unique fields should be configurable for authentication ?

jgirdner:
I accomplish this with a simple backend. Not sure if it needs to be added to this module. 

`
from django.conf import settings
from django.contrib.auth.models import User


class EmailOrUsernameModelBackend(object):
    def authenticate(self, username=None, password=None):
        if username:
            if '@' in username:
                kwargs = {'email': username}
            else:
                kwargs = {'username': username}
            try:
                user = User.objects.get(**kwargs)
                if user.check_password(password):
                    return user
            except User.DoesNotExist:
                return None
                

    def get_user(self, user_id):
        try:
            return User.objects.get(pk=user_id)
        except User.DoesNotExist:
            return None
`

mmarksnippety:
I read carefully this pull request. I think these changes are going in wrong direction. 

IMHO, the better way to deal with various authentication methods is  to extend or customize authentication backend. 

We want to keep authentication data formed in two fields: A public identifier (email, username, phone...) and A private identifier (password, token, TOTP...)

mmarksnippety:
After discussion in our company, we decide merge this pull request. We see advantages to make authentication process more flexible. I will deal with this in a few days

haxoza:
In general it is good idea and the code in pull request looks good. My only concerns are:

1. If we really should hardcode `'email'` field name in serializer and in new auth backend. For custom user models is can cause some issues.
2. Maybe instead of doing it we should check if can integrate some other libraries that provide such auth backends (like i.a. django-allauth does).
3. The auth backend in pull request doesn't handle the case when `UserModel.DoesNotExist` or `MultipleObjectsReturned` exceptions are raised on `UserModel.objects.get()` method call.

mmarksnippety:
In pull request #128 we introduce easy way to use custom serializers. I think is good way to make own login serializer that support email.

haxoza:
> I think it is good way to make custom login serializer that support email

:+1:

piotr-szpetkowski:
It's much simpler to use `USERNAME_FIELD` in custom `User` model to solve the issue. Besides that - `User` model does not have to contain field called `email` and it does not even have to contain any email field at all. Therefore I find this PR invalid.

Issue #60, opened 2015-07-13
Some url names clash with default Django url names.


mcastle:
For instance, login and logout clash with standard Django registration urls.

haxoza:
True, thanks! We didn't noticed that because we mainly deployed pure REST django apps with djoser.

konradhalas:
@haxoza we had offline discussion about this some time ago - can you remember me your thoughts?

mmarksnippety:
I make pull request #129. It introduces configurable URL namespace. We plan to merge it in this week

PR #69, opened 2015-08-04
Add namespace to djoser urls to prevent url clashes

mcastle:
Prevents url clashes with Django urls and other lib urls. This fixes issue #60.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Configure `urls.py`:
 ```python
 urlpatterns = patterns('',
     (...),
-    url(r'^auth/', include('djoser.urls')),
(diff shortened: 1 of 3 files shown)


lwm:
Any movement on this? Seems to make sense, as some of the names are quite general, for example, `register`, `user`, `login`.

mmarksnippety:
Adding namespace breaks compatibility. So we plan to add it in the next major release. First we release minro version with fixes to Django 1.9

lwm:
> So we plan to add it in the next major release.

:rocket:

mmarksnippety:
I make pull request #129. It introduce configurable url namespace.

PR #71, opened 2015-08-06
Fix deprecated DRF code

mcastle:
DRF 3.2.0 breaks djoser, because `request.DATA` is deprecated. See here: http://www.django-rest-framework.org/topics/3.2-announcement/


haxoza:
Good catch! Thanks!

mcastle:
For sure!

Issue #72, opened 2015-08-09
Empty POST request to /login causes an exception


optimistiks:
If you make a request to login endpoint without any data, let's say

```curl -X POST http://127.0.0.1:9000/auth/login/```

exception will occur

```
KeyError at /user/login/
'username'
Request Method:	POST
Request URL:	http://127.0.0.1:9000/auth/login/
Django Version:	1.8.2
Exception Type:	KeyError
Exception Value:	
'username'
Exception Location:	/dir/env/local/lib/python2.7/site-packages/djoser/serializers.py in validate, line 102
```

I think it's better to return response like this

```
{"username":["This field is required."],"password":["This field is required."]}
```

Issue #64, opened 2015-07-20
Return login user data in login call


s2krish:
Hay Guys,

Currently login function returns toke on successful authentication which is very nice. 

From the practical prospective, wouldn't it be nice to return user data? For example in a real app there's many API calls which uses user Id. But to get user user Id, another get request to /me has to be made. If login function returns user data + tokens , one request can be avoided.

Thanks

haxoza:
Sure, good idea. We will think about that.

alican:
Will it be available in the next version?

konradhalas:
Hi @s2krish thank you for your feedback and sorry that you have to wait so long.

We decided that each endpoint should be independent and should have only single responsibility. I understand that from the client point of view, it will be easier to receive user data after login in 99% cases, but for a purity reasons we are not doing it. This is why we also removed `LOGIN_AFTER_REGISTRATION` and `LOGIN_AFTER_ACTIVATION` flags (#57). Please check also #56.

As always, this is a trade off, but I hope our thinking is ok in this matter.

hammadzz:
This is such a common case. Is there any documentation how to be able to implement these easily? I also want to have login after registration functionality. This is probably what most people want to do, there should be documentation how to easily add such functionality.

nitrag:
Purity < User Experience.

PR #73, opened 2015-08-09
Fixed 2 issues

glothriel:
Return login user data in login call #64
Empty POST request to /login causes an exception #72


smcoll:
@glothriel thanks from another user.  We ran into this error as well, when the keys were not named correctly.

PR #79, opened 2015-08-29
Fix setup script to include urls package.

alekseyr:


haxoza:
Thanks for your proof of concept however we did it a little bit simpler.

PR #81, opened 2015-09-14
Fix write_only_fields deprecated in DRF 3.2

mjuopperi:
The write_only_fields was marked deprecated in 3.0 and deprecated in 3.2. In the current version the users hashed password is returned in the JSON when registering. This commit fixes that.

http://www.django-rest-framework.org/topics/3.2-announcement/#deprecations


haxoza:
Thanks!

PR #83, opened 2015-09-23
Fixed deprecated and added style for password

yiyocx:
Removed deprecated write_only_fields in DRF 3.2 and added style for the input password in the register browsable API


haxoza:
Awesome! Thanks!

PR #84, opened 2015-09-29
Include testproject in source distribution.

fladi:
This enables packagers like me to run the tests automatically during package
build.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
 include README.md
 include djoser/templates/*
 include requirements.txt
+recursive-include testproject *.py


konradhalas:
Hi @fladi - thank you for your PR!.

So we should also include `requirements-test.txt` - am I right?

piotr-szpetkowski:
It is not a common / good practice to include tests in python projects - especially such simple ones like djoser. Besides that - you could say that is rather redundant to run tests during bulding package for your distribution because:

1. We already do that - everything that is in master has been verified by CI.
2. Most people will use djoser via PyPI anyway, because it's more convenient to jail appropriate version in a project virtualenv instead of having a freeze on global version.

PR #86, opened 2015-10-06
 Add optional setting PASSWORD_VALIDATORS 

sh4r3m4n:
By default it's an empty list. When it has values, the password changes serializers validate the new password against the given validators.
Very useful with [django-passwords](https://github.com/dstufft/django-passwords/blob/master/passwords/validators.py) validators module.


sh4r3m4n:
I updated this to pass the tests with django<1.7

haxoza:
Nice, good idea! And thanks for your pull request. It looks good so we would be happy to merge it. 

However it would be good to add at least one test to our test suite that overrides the password validators setting.

sh4r3m4n:
Ok, I will code some tests the next days

mmarksnippety:
Thanks for your pull request. We plans to merge #117 that are derived from yours pull.

Issue #93, opened 2015-11-08
Add settings for easily swapping serializers.


Debetux:
Hi,

I wanted to know if it could be a good idea to add new settings for using custom serializers for each endpoint without overriding urls and views. I'm currently using djoser ; and since you removed `LOGIN_AFTER_*` settings, it's difficult to modify the library behaviour. I think theses settings could really be helpful.

What do you think?
If you think it worth the implementation, I'll make a PR in no time. :)

Thanks,
Lancelot.

konradhalas:
Hi @Debetux - thank you once again for your feedback.

Yes, it sounds like a good idea, but TBH I like current solution, it's clean and it does not require any additional mechanism (code). Less code == less bugs, but I understand that it can be a little bit inconvenient when you have to override views and provide custom urls.

So I will say that we can think about it. What other pieces should we configure in this way? This issue somehow related with #91.

@haxoza @rugala - any thoughts?

Debetux:
No problem.
I don't see any other pieces to configure this way right now.

Well, the problematic is that I want to add some custom behaviour without having to add too much complexity in my code (like adding a serializers field) / or for the client. Adding settings relating to serializers seemed to be the simplest solution.
Otherwise, may be there's others way to do this. :)

mmarksnippety:
Settings custom serializers are introduced in #128

Issue #91, opened 2015-11-03
possible E-mail improvement


apokinsocha:
I want to quickly say first Djoser is a great package to kickoff DRF REST API projects and I've used it to its absolute maximal intent and beyond. Great job!

I figure I would add my 2cents on some improvements though. Well, actually it's just one, the rest I found very flexible and inline with OOP/DRY.

With respect to sending e-mails, I would look into decoupling it from the DRF Views completely.
Say for example you have attachments (which is not so uncommon), with Djoser as is, I had to make some minor but ugly modifications to Djoser and email views I sub-classed.

Currently Djoser does something similar with Email as TemplateView (or more generally, class based views) in Django, in that there is a "get_context_data" method etc...that is passed in to render the text (the e-mail subject, body etc...) given the path to the template (subject.html, body.html etc...).

This isn't bad, in most cases, but again, if I have to deal with attachments, and also with different types of scenarios (send to multiple recipients, bccs etc...), this may not be the best solution.

Instead of coupling that to the DRF Views, I would decouple it completely and anytime Email functionality is needed, simply reference an Email class within the code and call it.

Here is simple email class I wrote (I called it mixin class by accident) I'm using currently in one of my projects with djoser:

https://github.com/apokinsocha/django-email-mixin/blob/master/django-email-mixin.py

It maybe a bit redundant with respect to djangos existing email function, but I wanted to encapsulate some additional common functionality with respect to e-mails.

Usage example:

``` python
    _file = some_file
    html_body = DjangoEmailWrapper.get_template_text('email_body.txt', context=context)
    subject = DjangoEmailWrapper.get_template_text('subject.txt', context=context, inline=True)
    emails = ['test@gmail.com', 'test2@gmail.com']
    msg = DjangoEmailWrapper(subject, getattr(settings, 'DEFAULT_FROM_EMAIL', None), html_body, bcc=emails, html_body=html_body)
    msg.attach_file(_file)
    msg.send_email()
```

I can do a PR in the near future to integrate it, if there is foreseeable use beyond my own.
A pretty simple change in my view, but I think it cleans things up a bit and also allows to easily call DjangoEmailWrapper in other non-djoser views.

mcgeeco:
I agree with this. I actually had to do something similar as well for doing things like sending activation emails after login attempts for inactive accounts.

apokinsocha:
sounds good. I'll see about a PR within the next few weeks.

konradhalas:
[...]

PR #95, opened 2015-11-12
More customisable user/registration serializers

charanpald:
I found it slightly complicated to allow additional fields for the user (e.g. first_name, last_name and fields in objects related by one-to-one keys). As far as I understood, this required adding REQUIRED_FIELDS to the User model which involves creating a custom user model. A simpler solution was just to allow user-defined serializers, which I implemented here (the defaults are UserSerializer and UserRegistrationSerializer). At the same time I allowed alternative locations for the activation email templates which saves overriding the registration view. 

There are also a couple of fixes in the commits:   
1. When SEND_ACTIVATION_EMAIL is true a newly created account should be inactive 
2. When the UID in activation is invalid, a simple error message is given (as opposed to the full detailed exception). 

Charanpal 

P.S. Thanks for the great library!

### DIFF
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -61,6 +61,8 @@ class RegistrationView(utils.SendEmailViewMixin, generics.CreateAPIView):
         signals.user_registered.send(sender=self.__class__, user=instance, request=self.request)
         if settings.get('SEND_ACTIVATION_EMAIL'):
             self.send_email(**self.get_send_email_kwargs(instance))
+            instance.is_active = False            
+            instance.save()
 
     def get_email_context(self, user):
         context = super(RegistrationView, self).get_email_context(user)
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -53,8 +53,8 @@ class RegistrationView(utils.SendEmailViewMixin, generics.CreateAPIView):
         permissions.AllowAny,
     )
     token_generator = default_token_generator
-    subject_template_name = 'activation_email_subject.txt'
-    plain_body_template_name = 'activation_email_body.txt'
+    subject_template_name = settings.get("ACTIVATION_EMAIL_SUBJECT")
+    plain_body_template_name = settings.get("ACTIVATION_EMAIL_BODY")
 
     def perform_create(self, serializer):
         instance = serializer.save()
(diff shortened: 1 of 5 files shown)


djoser/views.py
```diff
@@ -46,19 +48,21 @@ class RegistrationView(utils.SendEmailViewMixin, generics.CreateAPIView):
...
-    serializer_class = serializers.UserRegistrationSerializer
+    serializer_class = registration_serializer_class
     permission_classes = (
         permissions.AllowAny,
     )
     token_generator = default_token_generator
-    subject_template_name = 'activation_email_subject.txt'
-    plain_body_template_name = 'activation_email_body.txt'
+    subject_template_name = settings.get("ACTIVATION_EMAIL_SUBJECT")
+    plain_body_template_name = settings.get("ACTIVATION_EMAIL_BODY")
```
mcgeeco: Does anyone else think that while this is being done, it would be worth adding a ACTIVATION_EMAIL_BODY_HTML as well?

konradhalas:
Hi @charanpald, thank you for your PR - good job!

It looks like many people have similar problem and they found extending djoser not so easy - your pull request is related to issue #93 and (partially) to issue #91. I have to say, that now I'm convinced - we should definitely introduce something like this.

However, I see a few things which we have to improve in your PR before merge:

1. We should support overriding of all serializers - I know that probably user and registration serializers are the most useful here, but just for consistency we should allow to configure them all. 
2. Your PR failed with Python 2.6, we should handle this somehow (drop support or fix your code).
3. It will be awesome if you can add some tests.
4. Configuration of email templates and invalid UID error should not be part of this PR - feel free to open a new one.

Let me know if you are planing to improve your PR or should I handle it on my own.

djoser/views.py
konradhalas: Yes, definitely.

charanpald:
Hi @konradhalas, 
I will try to find some time to write test cases for the serializers in the next few days. As for overriding all serializers (apart from user and registration), I am not sure what the use cases for doing that are - can you give some examples of those please? 

I will remove the stuff about email templates and invalid UID errors and put it in another PR. My personal thoughts on supporting Python 2.6 is that it is not required (it's relatively easy to upgrade to 2.7 which has been around for a while). 

Charanpal

mmarksnippety:
We introduce custom serializers in #128 (Inspirated by Your pull request). Human readable invalid UID is introduced in #127. And SEND_ACTIVATION_EMAIL behavior is fixed in #126.

PR #100, opened 2015-12-01
Convert User.DoesNotExist to string before passing to ValidationError

carlosfunk:
If I post an invalid uid to the activation endpoint it passes the DoesNotExist exception directly to serializers.ValidationError which is raises `TypeError: DoesNotExist('User matching query does not exist.',) is not JSON serializable` and returns a 500 error to the user instead of a 400.


haxoza:
Thanks @carlosfunk! Good catch!

PR #104, opened 2015-12-11
djoser returns a HTTP 400 for a stale activation token

lwm:
Hi folks, I assume this should be standard behaviour so I went ahead and wrote it. 

Let me know what you think :+1:

### DIFF
--- a/djoser/constants.py
+++ b/djoser/constants.py
@@ -3,6 +3,7 @@ from django.utils.translation import ugettext_lazy as _
 INVALID_CREDENTIALS_ERROR = _('Unable to login with provided credentials.')
 INACTIVE_ACCOUNT_ERROR = _('User account is disabled.')
 INVALID_TOKEN_ERROR = _('Invalid token for given user.')
+STALE_TOKEN_ERROR = _('Stale token for given user.')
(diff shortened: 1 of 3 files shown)


haxoza:
Thanks! At first glance it looks good.

Do you think that it is really necessary to detect situation when the activation URL was reached for the second time?

lwm:
Cool, happy to help.

As for as I understand it, firing a `POST` request at the `/activate/` end-point is idempotent. However, in terms of user experience, a reason could be that if the user clicks the link and something bad happens (browser drops, tab closes, etc.) and they are not sure what the result was, they can try to click the activation link again and they will be told that it is invalid.

lwm:
Oh, and as noted by a colleague, perhaps returning a `419` response would be better, as noted in [#158](https://github.com/GetBlimp/django-rest-framework-jwt/issues/158#issuecomment-161930536).

haxoza:
@lwm Thanks for your work. I've just merged the PR but we decided that 403 response status code is more suitable in this situation. If 419 would be used for libraries like DRF-JWT I can imagine that it could interfere with authentication mechanisms on the client side.

However after long discussion we are still not sure if the 400 code is just even better.

lwm:
@haxoza Sounds good :rocket:

Issue #40, opened 2015-04-26
Add JSON Web Token authentication support


haxoza:
It would great to support the following project that implements JWT auth backend: https://github.com/GetBlimp/django-rest-framework-jwt

konradhalas:
+1 Any ideas how we could support DRF token, our custom multi device token and JWT? Probably we should implement some generic layer between our views and authentication backends.

avimeir:
I'm already using djoser with DRF-JWT and it seems to be working just fine. Is it a weird unexpected feature?

haxoza:
Sounds awesome! It would be cool to have it documented and tested. @avimeir, we will appreciate any pull requests in that matter!

avimeir:
Try to turn on token-based authentication on DRF (http://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication) and then follow the usual install of DRF-JWT, then added djoser and it works right of the box with no modifications to the code.

NabeelValapra:
@avimeir, 
1. If so still djoser will be creating token in the database? how did you overcame it?
2. Will be there any change in the flow of /login/,?

avimeir:
@NabeelValapra 
1. I assume that djoser will use the authentication class configured by DRF, which happens to be DRF-JWT (rest_framework_jwt.authentication.JSONWebTokenAuthentication). I must admit that I haven't looked into it, so the fact that it's working is pure magic at this point :)
2. Well, you need to follow the DRF-JWT flow for issuing a JWT, by default the endpoint is: ^api-token-auth/

liaoliaopro:
+1 to support JWT. make authentication backend configurable instead of binding with Token auth.

avances123:
I have read the @haxoza POC in this matter and I think is the way to do it, by the way:

@NabeelValapra 
1.  You use rest_framework_jwt.authentication.JSONWebTokenAuthentication, and delete 'rest_framework.authtoken' from your installed apps. At this point, /auth/login/ endpoint is broken (maybe some others) as the token table is not created
2. I override djoser login url with drf-jwt 
```
url(r'^auth/login/', 'rest_framework_jwt.views.obtain_jwt_token'),
```

cerdman:
+1

uniqueg:
What is the status here? Did anyone get it to work, reproducibly? Would be much appreciated...

konradhalas:
Hi @uniqueg, you can always try something like this:

```python

import rest_framework_jwt.views
[...]

PR #105, opened 2015-12-12
Demonstrate jwt auth support

lwm:
Following from [this comment](https://github.com/sunscrapers/djoser/issues/40#issuecomment-96586800) on #40, here are the docs/test. I see you're working on something new with #33 but this should be a help to people in the short term.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,18 @@ Run migrations (if you are using Django 1.7+ or South) - this step will create t
 
     $ ./manage.py migrate
 
+##### Note.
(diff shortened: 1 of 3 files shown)


lwm:
Ugh, the error is to do with the underlying `pyjwt` library blowing up in the `python3.2` environment. It appears to not be supported by them. Oh well, I don't suppose it is a big deal, feel free to close since this was only a 'nice to have'.

mcgeeco:
I've integrated rest-framework-jwt with djoser but need to copy the code to Github (we use Stash). Hopefully when I get time I can go through the steps I took and hopefully that'll be of some use.

lwm:
Closing this one off. I am sure the maintainer will take care of this in due time.

Issue #114, opened 2016-01-29
SEND_ACTIVATION_EMAIL doesn't work correct


slav0nic:
With current implementation User created with `is_active=True` even `SEND_ACTIVATION_EMAIL` set

lwm:
FYI - #126 will resolve this :+1:

slav0nic:
tnx

PR #115, opened 2016-01-29
re #114: create inactive user if SEND_ACTIVATION_EMAIL set

slav0nic:


mmarksnippety:
Thank You for pull request. We plan merge #126 which is based on your code.

PR #116, opened 2016-02-03
Added warning for referencing User directly from django.contrib.auth.models

yonkornilov:
Changed readme to help prevent a mistake which would result in "AttributeError: Manager isn't available; User has been swapped for \'app.UserModel\'"

If one was to use the web api from django-rest-framework as is, install Djoser, then try to implement a custom user model, they would run into this issue. Though I'm very new to Django and perhaps other users would not run into this problem, I had a lot of trouble with this since most of my searching for the said error lead to old StackOverflow questions that were unrelated to Djoser.

I think adding this little bit of text could potentially save users of Djoser from a lot of time searching online and looking in the wrong direction.


yonkornilov:
I was quite tiered when I made this and didn't write my change to the README in the best way possible, I'll just open a suggestion on the repo instead

yonkornilov:
Also this is outside the scope of Djoser and is more related to django-rest-framework, I just realized

PR #123, opened 2016-02-24
Django 1.9 compatibility

mmarksnippety:
Improve TOX configuration. Now, dependencies are specified by a version range.
Rather than a version specified by a patch revision.
We don't need to commit anything when new patch revision becomes available.
To our taste, this is fair - testing all historical patches takes a lot of time
and end-users should use the newest revision always.

Remove deprecated django.conf.urls.patterns
This function is deprecated since Django 1.8. We remove usage of this
function for all supported Django versions. For more detail read:
https://docs.djangoproject.com/en/1.8/ref/urls/#patterns

Remove TEMPLATE_DIRS from testproject settings. This settings is deprecated in Django 1.8.
We remove this setting, because is not used in testproject, and is deprecated in Django 1.8.
This setting was required in Django 1.4 and not in Django 1.5.
Read more at: https://docs.djangoproject.com/en/1.8/ref/settings/#template-dirs

Travis configured to use python 3.2
Travis uses by default virtualenv 14. This version of virtualenv do not support Python 3.2.
So we configure Travis to install virtualenv in version lower than 14.
Read more about this issue: https://github.com/travis-ci/travis-ci/issues/5485
We will come back to newest virtualenv, when we stop supporting Python 3.2

Added TEMPLATES settings to testproject for Django >= 1.8.
This setting will be introduced in Django 1.8, and it will be obligatory in Django 1.10.
The lack of this setting in Django 1.9 causes warning

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ env:
   - TOX_ENV=py35-dj19-drf33
 
 install:
-  - pip install tox
+  - travis_retry pip install "virtualenv<14.0.0" "tox>=1.9"
 
 script:
   - tox -e $TOX_ENV
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ Supported Django versions:
  * Django 1.6
  * Django 1.7
  * Django 1.8
+ * Django 1.9
 
 Supported Django Rest Framework versions:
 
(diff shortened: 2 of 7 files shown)


haxoza:
Looks good! Let's merge it!

PR #125, opened 2016-02-24
Use RegexField in UidAndTokenSerializer.

tyler46:
Hi, 

I have changed the two fields ( `uid` and `token` ) in `UidAndTokenSerializer` since these fields are expected to meet some regex. I have just ported from django's  auth urls.
Perhaps you like it.

Thank you

### DIFF
--- a/djoser/serializers.py
+++ b/djoser/serializers.py
@@ -62,8 +62,8 @@ class PasswordResetSerializer(serializers.Serializer):
 
 
 class UidAndTokenSerializer(serializers.Serializer):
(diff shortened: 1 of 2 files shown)


mmarksnippety:
Hello. 
We don't want to change this fields types. `RegexField` generate error messages that are unclear for end user. This pull request https://github.com/sunscrapers/djoser/pull/127 provide human readable error message when incorrect uid is provided

tyler46:
Hi @mmarksnippety sure no problem.

PR #126, opened 2016-02-25
Set user activity to false if we fire an activation email

### LINKED ISSUE #114: SEND_ACTIVATION_EMAIL doesn't work correct

lwm:
Haven't seen any activity on #115, and it should definitely get fixed, so here is the change and a test.

Closes #114.

### DIFF
--- a/djoser/serializers.py
+++ b/djoser/serializers.py
@@ -33,7 +33,11 @@ class UserRegistrationSerializer(serializers.ModelSerializer):
         )
 
     def create(self, validated_data):
-        return User.objects.create_user(**validated_data)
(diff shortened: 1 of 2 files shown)


mmarksnippety:
Thank You for this pull request! It will be merged soon. I plan new release tomorrow with some fixes.

haxoza:
Looks reasonable to have it in djoser. If somebody uses just built-in Django user this change is really helpful.

lwm:
> Thank You for this pull request! It will be merged soon.
:hammer:

PR #127, opened 2016-02-25
Human readable error message when incorrect uid is provided

mmarksnippety:
We add invalid uid error message to UidAndTokenSerializer.
This serializer is used in ActivationView and PasswordResetConfirmView.


haxoza:
Looks good! Let's squash the commits and merge it.

PR #128, opened 2016-02-26
Introduced DJOSER.SERIALIZERS

mmarksnippety:
You can use own serializers by simple change settings.
We add SerializerFromSettingsMixin to utils module.
This mixin load serializer class from settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,8 @@ DJOSER = {
     'PASSWORD_RESET_CONFIRM_URL': '#/password/reset/confirm/{uid}/{token}',
     'ACTIVATION_URL': '#/activate/{uid}/{token}',
     'SEND_ACTIVATION_EMAIL': True,
-    'PASSWORD_VALIDATORS': []
+    'PASSWORD_VALIDATORS': [],
+    'SERIALIZERS': {},
 }
 ```
 
@@ -456,6 +457,35 @@ These validators are run on `/register/` and `/password/reset/confirm/`.
 
 **Example**: `[my_validator1, my_validator2]`
 
+### SERIALIZERS
+
+This dictionary is used to update the defaults, so by providing, 
+let's say, one key, all the others will still be used.
+
+**Examples**
+```
+{
+    'user': 'myapp.serializers.SpecialUserSerializer',
+}
+```
+
+**Default**: 
+```
+{
+    'activation': 'djoser.serializers.ActivationSerializer',
+    'login': 'djoser.serializers.LoginSerializer',
+    'password_reset': 'djoser.serializers.PasswordResetSerializer',
+    'password_reset_confirm': 'djoser.serializers.PasswordResetConfirmSerializer',
+    'password_reset_confirm_retype': 'djoser.serializers.PasswordResetConfirmRetypeSerializer',
+    'set_password': 'djoser.serializers.SetPasswordSerializer',
+    'set_password_retype': 'djoser.serializers.SetPasswordRetypeSerializer',
+    'set_username': 'djoser.serializers.SetUsernameSerializer',
+    'set_username_retype': 'djoser.serializers.SetUsernameRetypeSerializer',
+    'user_registration': 'djoser.serializers.UserRegistrationSerializer',
+    'user': 'djoser.serializers.UserSerializer',
+}
+```
+
 ## Emails
 
 There are few email templates which you may want to override:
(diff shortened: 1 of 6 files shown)


djoser/utils.py
```diff
@@ -8,6 +8,16 @@
 except ImportError:
     from django.contrib.sites.models import get_current_site
 
+try:
+    from django.utils.module_loading import import_string
+except ImportError:
+    from django.utils.module_loading import import_by_path
```
Stranger6667: `from django.utils.module_loading import import_by_path as import_string` will be simpler, than another function

mmarksnippety: Good point!

README.md
```diff
@@ -549,9 +579,10 @@ In this extremely short tutorial we are going to mimic the simplest flow: regist
 
 ## Customization
 
-If you need to override some `djoser` behaviour, you could define your custom view/serializer.
+If you need a use custom serializer, use `DJOSER.SERIALIZERS`.
```
haxoza: Shouldn't be `DJOSER['SERIALIZERS']`?

README.md
```diff
@@ -456,6 +457,35 @@ These validators are run on `/register/` and `/password/reset/confirm/`.
 
 **Example**: `[my_validator1, my_validator2]`
 
+### SERIALIZERS
+
+This dictionary be used to update the defaults, so by providing,
```
haxoza: > This dictionary be used to update the defaults

should be

> This dictionary **is** used to update the defaults

README.md
```diff
@@ -549,9 +579,10 @@ In this extremely short tutorial we are going to mimic the simplest flow: regist
 
 ## Customization
 
-If you need to override some `djoser` behaviour, you could define your custom view/serializer.
+If you need a use custom serializer, use `DJOSER[SERIALIZERS]`.
+And if you need to override some `djoser` behaviour, you could define your custom view/serializer.
```
haxoza: I would change it to something like this:

If you need to customize some serializers behavior you can use `DJOSER[SERIALIZERS]` setting to use your own serializer classes in the built-in views. Or if you need to completely change the default `djoser` behaviour you can always override `djoser` views by your own custom ones.

mcgeeco: Or even:
If you need to customize any serializer behavior you can use the DJOSER['SERIALIZERS'] setting to use your own serializer classes in the built-in views. Or if you need to completely change the default djoser behaviour, you can always override djoser views with your own custom ones.

testproject/testapp/tests.py
```diff
@@ -616,3 +625,73 @@ def test_put_should_update_user(self):
...
+        serializers_manager = djoser.serializers.SerializersManager({
+            'user': djoser.serializers.UserSerializer
+        })
+        self.assertRaises(Exception, serializers_manager.get, 'bad_name')
+
+    def test_get(self):
+        serializers_manager = djoser.serializers.SerializersManager({
+            'user': djoser.serializers.UserSerializer})
+        serializer_class = serializers_manager.get('user')
+        self.assertTrue(issubclass(serializer_class, djoser.serializers.UserSerializer))
```
haxoza: Why is this exactly the same as `test_get_serializer()`?

mmarksnippety: Good point. Is exactly the same, because of refactoring SerializeManager. Now it is redundant. I'll remove one of this tests.

testproject/testapp/tests.py
```diff
@@ -616,3 +625,73 @@ def test_put_should_update_user(self):
...
+            self.assertFalse(load_serializer_mock.called)
+
+    def test_get_serializer_class_not_exists(self):
+        serializers_manager = djoser.serializers.SerializersManager({})
+
+        if django.VERSION >= (1, 7):
+            self.assertRaises(ImportError, serializers_manager.load_serializer, 'bad.UserSerializer')
+        else:
+            from django.core.exceptions import ImproperlyConfigured
+            self.assertRaises(ImproperlyConfigured, serializers_manager.load_serializer, 'bad.UserSerializer')
```
haxoza: I'm not sure for what reason is this test. Can you clarify it?

README.md
haxoza: @mcgeeco true, thanks! :+1:

haxoza:
@mmarksnippety I really like the general idea behind `SerializersManager`. Let's fix last small issues and merge it! :tada:

testproject/testapp/tests.py
mmarksnippety: In fact, this test case shows that we can't load not existing class. This is evident in all. We do not support this exception in any way. So we don't need this test.

PR #129, opened 2016-03-01
Add urls namespace as option

mmarksnippety:


testproject/testapp/tests.py
```diff
@@ -27,6 +29,9 @@ class RootViewTest(restframework.APIViewTestCase,
                    assertions.StatusCodeAssertionsMixin):
     view_class = djoser.views.RootView
 
+    def tearDown(self):
+        clear_url_caches()
```
haxoza: @mmarksnippety can you leave short comment in a code why it is here? I guess that django urls needs some internal clean-up after changing `ROOT_URLCONF` in `test_get_should_return_urls_mappings_whit_namespace_setup()` - am I right?

piotr-szpetkowski:
Hi. Unfortunately, this PR is no longer valid since the `RootView` has been completely rewritten. Also every other part of djoser that this PR touches has also changed more or less.

Therefore I'm going to close this, however thank you very much for your work.

PR #133, opened 2016-03-17
Fix/fixing reset password

nnkps:
Returning BAD REQUEST code instead of OK for reset password view when there is no user with given email.

btw. Case with test passing on pair programming had to do with generator in get_users method. "if not users" in validate_email method in ResetPasswordSerializer wasn't triggered because there was generator object inside of a variable.


haxoza:
Awesome! Goog job! :+1: :tada:

Could you squash the commits and get rid of version bump, please? I think this will be merged into 0.5.0 version because it is not backward compatible and it changes the API behaviour a little bit.

hvdklauw:
Ehh that's actually really bad, you want to give back a 200 OK when the email doesn't exist, that way people can't use it figure out of a certain e-mail address is available in the system. (You basically give them half the login credentials in case of email/password authentication)

m-nako:
I agree with @hvdklauw that leaking the information whether a certain email exists in the database is a bad idea. At the same time, I have seen client requests for the error message stating whether a reset password email has been sent or not.

I'd suggest adding a setting to enable that feature but keep the safer option as a default and document that thoroughly.

m-nako:
Merged @nnkps 's changes (thanks!) and added `PASSWORD_RESET_SHOW_EMAIL_NOT_FOUND` setting.

PR #135, opened 2016-03-25
Implement LOGOUT_ON_PASSWORD_CHANGE setting

### LINKED ISSUE #113: Logout after password change
As of django >= 1.7 setting a new password has the following new behaviour (From [Django source] (https://github.com/django/django/blob/1.8.8/django/contrib/auth/views.py#L304)):

```python
# Updating the password logs out all other sessions for the user
# except the current one if
# django.contrib.auth.middleware.SessionAuthenticationMiddleware
# is enabled.
update_session_auth_hash(request, form.user)
```

It would be nice to adapt the code to address this issue. Maybe a new setting `LOGOUT_ON_PASSWORD_CHANGE ` would make sense to make use of this new behaviour.

Here some similar projects which already handle this:

- [django-rest-auth](https://github.com/Tivix/django-rest-auth/blob/master/rest_auth/serializers.py#L258)
- [django-user-accounts](https://github.com/PersonalGenomesOrg/django-user-accounts/blob/8d43830ace662802d5071454b701fe96a94e9903/account/views.py#L472)

buoto:
Closes #113

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@
 .coverage
 dist/
 testproject/db.sqlite3
+
+# Vim swap files
+[._]*.s[a-w][a-z]
+[._]s[a-w][a-z]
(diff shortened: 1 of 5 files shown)


m-nako:
Thank you, @buoto! I've added info on the setting to readme and pulled the request.

PR #139, opened 2016-04-28
Make Token serializer customizable

asennikov:
If it doesn't break anything, I'd like to have this as option.


haxoza:
Thanks! That's good catch! 👍 

Merged in c4d26bdf4496bfd32cb2719fd24620a754a1c8fd! 🎉

PR #141, opened 2016-05-18
Refactor SendEmailMixin. 

serathius:
Replacing mixin with email factory. 
Direction for enhancement for #91.
Possibly email factories deserve own module.

### DIFF
--- a/djoser/utils.py
+++ b/djoser/utils.py
@@ -1,7 +1,9 @@
 from django.conf import settings as django_settings
+from django.contrib.auth.tokens import default_token_generator
 from django.core.mail import EmailMultiAlternatives, EmailMessage
 from django.template import loader
 from rest_framework import response, status
+from . import settings
 
 try:
     from django.contrib.sites.shortcuts import get_current_site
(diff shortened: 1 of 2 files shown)


djoser/utils.py
```diff
@@ -62,48 +44,74 @@ def post(self, request):
...
-        return self.subject_template_name
+class UserActivationEmailFactory(UserEmailFactoryBase):
+    subject_template_name = 'activation_email_subject.txt'
+    plain_body_template_name = 'activation_email_body.txt'
 
-    def get_plain_body_template_name(self):
-        return self.plain_body_template_name
+    def get_context(self):
+        context = super(UserActivationEmailFactory, self).get_context()
+        context['url'] = settings.get('ACTIVATION_URL').format(**context)
```
guillaumevincent: why don't you use django_settings ? You should don't use your own settings : 


> Also note that your code should not import from either global_settings or your own settings file. django.conf.settings abstracts the concepts of default settings and site-specific settings; it presents a single interface. It also decouples the code that uses settings from the location of your settings.

https://docs.djangoproject.com/en/dev/topics/settings/#using-settings-in-python-code

guillaumevincent:
Hi, why the use of djoser settings file and django_settings ?

> Also note that your code should not import from either global_settings or your own settings file. django.conf.settings abstracts the concepts of default settings and site-specific settings; it presents a single interface. It also decouples the code that uses settings from the location of your settings.

https://docs.djangoproject.com/en/dev/topics/settings/#using-settings-in-python-code

haxoza:
Hi @guillaumevincent! Thanks for pointing it out!

To be honest you're right. We probably should use `django.conf.settings` as an interface to the configuration. 

However we do some magic in for two reasons:

1. To shorten the amount of code (`settings.get('XXX')` instead of `settings.DJOSER['XXX']`)
2. To merge default settings with user specific settings (configurable serializers need that especially)

I think your finding is a candidate for separate issue. It does not only refer to this PR but to the whole djoser codebase.

It would be great if you can submit new issue about that.

Cheers!

m-nako:
I second @guillaumevincent's opinion about using the `django.conf.settings` object. We will need to open another issue to refactor that everywhere in the codebase, as @haxoza mentioned.

I've just merged this request, thank you @serathius!

PR #144, opened 2016-05-24
Simplified mixin, use directly function to send emails

### LINKED ISSUE #91: possible E-mail improvement

mrKisaoLamb:
close #91

### DIFF
--- a/djoser/utils.py
+++ b/djoser/utils.py
@@ -63,47 +63,32 @@ class ActionViewMixin(object):
 
 
 class SendEmailViewMixin(object):
-    token_generator = None
     subject_template_name = None
     plain_body_template_name = None
     html_body_template_name = None
-
-    def send_email(self, to_email, from_email, context):
-        send_email(to_email, from_email, context, **self.get_send_email_extras())
+    token_generator = None
 
     def get_send_email_kwargs(self, user):
         return {
             'from_email': getattr(django_settings, 'DEFAULT_FROM_EMAIL', None),
             'to_email': user.email,
             'context': self.get_email_context(user),
+            'subject_template_name': self.subject_template_name,
+            'plain_body_template_name': self.plain_body_template_name,
+            'html_body_template_name': self.html_body_template_name,
         }
-
-    def get_send_email_extras(self):
-        return {
-            'subject_template_name': self.get_subject_template_name(),
-            'plain_body_template_name': self.get_plain_body_template_name(),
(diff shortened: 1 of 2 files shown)


piotr-szpetkowski:
Hi, unfortunately this PR is not valid anymore since some time already and after giving it some thought I've decided it would be best to close it. I'm sorry it wasn't reviewed faster and thank you for your work.

PR #148, opened 2016-05-26
User creation should be atomic

pik:
Currently if the server or database connection is dropped between `User.objects.create_user(**validated_data)` and `user.save(update_fields=['is_active'])` - there will be an account created with `is_active=True` - this shouldn't be allowed.


m-nako:
True! Thanks a lot for that!

PR #150, opened 2016-05-26
Sanitization for username + tests. Closes #146

### LINKED ISSUE #146: Add sanitization option to username in login and registration serializers
It's a common pattern to store usernames and/or email addresses in a database for example in lower cases. Once username is stored in lower cases then incoming username in POST data should be lowered as well in order to compare them properly.

The only questions here is about good practices. Should it be possible to sanitize username in serializers or should it be done in user model and in Django authentication backend?

ppeczek:
Didn't work out sanitization for dictionary, because LoginSerializer doesn't provide dictionary. Could go around, but decided to create sanitization for string argument.
Included sanitization in 3 classes: UserRegistrationSerializer, LoginSerializer and SetUsernameSerializer.
Created 3 tests for each use case - similar to primary tests. Created list of 2 test sanitizers: uppercase and trim3, to check if works for list of sanitizers.

### DIFF
--- a/testproject/testapp/tests.py
+++ b/testproject/testapp/tests.py
@@ -666,6 +666,7 @@ class SetUsernameViewTest(restframework.APIViewTestCase,
         user = utils.refresh(user)
         self.assertNotEqual(user.username, data['new_username'])
 
+
 class UserViewTest(restframework.APIViewTestCase,
                    assertions.StatusCodeAssertionsMixin):
     view_class = djoser.views.UserView
--- a/testproject/testapp/tests.py
+++ b/testproject/testapp/tests.py
@@ -3,6 +3,7 @@ from django.conf import settings
 from django.contrib.auth import get_user_model, user_logged_in, user_login_failed, user_logged_out
 from django.contrib.auth.tokens import default_token_generator
 from django.core import mail
+from django.db import IntegrityError
 from django.test.utils import override_settings
 from django.test.testcases import SimpleTestCase
 from djet import assertions, utils, restframework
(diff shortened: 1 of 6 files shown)


ppeczek:
Not RTM, I will send commit with better coverage

haxoza:
@ppeczek in general it looks good. The coverage is decreased because tests fail on on Django <= 1.6.

Also I would add one more test in order to be sure that the proposed code works like it should. This test should check if registration view detects username uniqueness when sanitisers are applied.

ppeczek:
Sounds reasonable. Also requires users to apply their sanitizers on database. I will push another commit with test described by you and info in readme for users

ppeczek:
If description in README is clear, then RTM.

testproject/testapp/tests.py
```diff
@@ -101,6 +116,26 @@ def test_post_should_not_create_new_user_if_username_exists(self):
...
+            'username': 'jOhn',
+            'password': 'secret',
+            'csrftoken': 'asdf',
+        }
+        request = self.factory.post(data=data)
+        response = self.view(request)
+        self.assert_status_equal(response, status.HTTP_201_CREATED)
+
+        data = {
+            'username': 'joHn',
```
ppeczek: This shows dark side of sanitization. For uppercase/lowercase jOhn == joHn

piotr-szpetkowski:
As much as I admire your great work I think this stuff goes very broadly outside of djoser scope, which has a goal to deliver simple implementation of reusable and customizable common endpoints. I think it would be much better to push sanitizers as PR to DRF or to create a new package called `drf-sanitizers` or something alike.

Also https://groups.google.com/forum/#!topic/django-rest-framework/e-yxZx25_uE is worth looking at.

ppeczek:
I agree with you

28.07.2017 7:44 PM "Piotr Szpetkowski" <notifications@github.com>
napisał(a):

PR #152, opened 2016-06-02
Fixes #145: Added a new test for Django 1.10a1 and skipped the failing one if Django 1.10a1

### LINKED ISSUE #145: Test fails for Django 1.10a1 - default django auth backend treats credentials of inactive user as wrong ones
To reproduce install latest django and run:

`pip install -U django==1.10a1`

and then run tests.

martarozek:
It turns out that in Django 1.10a1 authenticate() returns None if the user is inactive. In previous versions it returned the authenticated user and it required an additional check (user.is_active) to determine whether or not the user can be logged in. So perhaps the change is not a bad thing and from Django 1.10a1 failed authentication could simply be communicated to the user as 'Bad credentials or account inactive'.

### DIFF
--- a/testproject/testapp/tests.py
+++ b/testproject/testapp/tests.py
@@ -1,3 +1,4 @@
+import unittest
 import django
 from django.conf import settings
 from django.contrib.auth import get_user_model, user_logged_in, user_login_failed, user_logged_out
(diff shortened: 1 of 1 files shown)


haxoza:
Awesome! Thanks a lot your work! 👍

m-nako:
Rebased to master. Thanks a lot!

PR #157, opened 2016-07-21
Send activation email whenever it changes. Closes #156

### LINKED ISSUE #156: User update should trigger an activation when email is changed
Hey guys, I think that it's a sane idea to check for email changes at the `UserView`. A change should trigger a new activation email. I can work on a PR, what do you think?

loop0:


loop0:
I'm gonna try to write some tests to improve coverage

djoser/serializers.py
```diff
@@ -24,6 +24,16 @@ class Meta:
             User.USERNAME_FIELD,
         )
 
+    def update(self, instance, validated_data):
+        email = instance.email
+        with transaction.atomic():
+            instance = super(UserSerializer, self).update(instance, validated_data)
+            if validated_data.get('email') and email != instance.email and settings.get('SEND_ACTIVATION_EMAIL'):
```
piotr-szpetkowski: I'd swap the order of  these conditions to avoid doing redundant conditions
``` python
settings.get('SEND_ACTIVATION_EMAIL') and validated_data.get('email') and email != instance.email
```

djoser/views.py
```diff
@@ -236,3 +236,16 @@ class UserView(generics.RetrieveUpdateAPIView):
 
     def get_object(self, *args, **kwargs):
         return self.request.user
+
+    def perform_update(self, serializer):
+        email = self.get_object().email
+        user = serializer.save()
+        signals.user_registered.send(sender=self.__class__, user=user, request=self.request)
+        if email != user.email:
+            if settings.get('SEND_ACTIVATION_EMAIL'):
```
piotr-szpetkowski: I suppose we could flatten those two conditionals:
``` python
settings.get('SEND_ACTIVATION_EMAIL') and email != user.email
```

haxoza:
In general that was good idea but now the content of activation email is misleading in this case.

It says ["You're receiving this email because you **created** an account on ..."](https://github.com/sunscrapers/djoser/blob/master/djoser/templates/activation_email_body.txt#L2). When you update an email you in fact have your account already **created**.

piotr-szpetkowski:
True, nice catch 👀

djoser/serializers.py
```diff
@@ -24,6 +24,16 @@ class Meta:
             User.USERNAME_FIELD,
         )
 
+    def update(self, instance, validated_data):
+        email = instance.email
```
haxoza: What will happen if `User` model does not have `email` attribute at all?

djoser/serializers.py
haxoza: This code is invalid for `User` model with `User. USERNAME_FIELD = 'email'` because the code inside `if` clause will never execute due to `read_only_fields` on serializer set to `User.USERNAME_FIELD`.

What's more in the use case mentioned in previous sentence the code for resending activation email should be present in `SetUsernameSerializer`.

PR #158, opened 2016-08-02
Password reset with frontend url validation for setting email domain

rwrzesien:
I have added a solution for issue with sending password reset email with correct domain in password reset url while having several frontend instances calling single API backend.

I have added PASSWORD_RESET_ALLOWED_HOSTS dict to settings where keys are urls to password reset pages (ie. 'http://www.sample.com/password_reset') and values are domain which will be used in buliding url in emails (ie. 'www.sample.com'). 

When request to reset password is sent from frontend, it should have 'frontend_url' parameter which is one of the keys from that setting, so it will set related domain in password reset url in email .


dekoza:
This is aimed at legacy version. If you need this for djoser-2.0, please add new pull request.

PR #159, opened 2016-08-12
Upon login notify user that the username attempted to login with does not exist

xtrinch:
Notify user that user with the username attempted to login with does not exist.

I think it would be great if there would be an option to notify the user that he is trying to login with an account that does not exist.

### DIFF
--- a/djoser/serializers.py
+++ b/djoser/serializers.py
@@ -55,6 +55,7 @@ class LoginSerializer(serializers.Serializer):
     default_error_messages = {
         'inactive_account': constants.INACTIVE_ACCOUNT_ERROR,
         'invalid_credentials': constants.INVALID_CREDENTIALS_ERROR,
+        'account_does_not_exist': constants.ACCOUNT_DOES_NOT_EXIST
     }
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


djoser/serializers.py
```diff
@@ -69,6 +70,8 @@ def validate(self, attrs):
                 raise serializers.ValidationError(self.error_messages['inactive_account'])
             return attrs
         else:
+            if User.objects.filter(**{User.USERNAME_FIELD: attrs.get(User.USERNAME_FIELD)}).first() is None:
```
piotr-szpetkowski: A better condition to check for would be:
`not User.objects.filter(**{User.USERNAME_FIELD: attrs.get(User.USERNAME_FIELD)}).exists()`

piotr-szpetkowski:
Also, we should probably pull it under a setting feature flag.

jasonrfarkas:
I'm not a security expert. (I was just was looking for examples of how others handled drj api for users.) But, I believe the suggested feature is actually a security concern. In general when a user doesn't have permission to view a resource the best practice is to just tell them it does not exist. You don't want to let them know that things exist but they can't access them, even more so with users.

piotr-szpetkowski:
Actually it seems like an author does not express an interest in further development of this feature. Therefore I'm going to close it now, however if it turns out that someone needs it than he/she is welcome to re-open it under new PR.

Issue #110, opened 2016-01-07
Updating djoser to a new version updates its dependencies


haxoza:
For example updating djoser to latest version when using django 1.7 updates django as well to latest version. It's because of `install_requires` in `setup.py` which is set to value of `requirements.txt` file. `install_requires` should be empty like e.g. in Django Rest Framework itself.

I believe that fix of this issue should update readme file too.

@bartekbrak thanks for your findings!

lwm:
Possibly worth looking into [pip-tools](https://github.com/nvie/pip-tools) for dependency management.

mmarksnippety:
@bartekbrak will do some more investigation about requirements

PR #167, opened 2016-09-27
Fixed required Django version

Stranger6667:
As mentioned in README - minimum Django version is 1.7


haxoza:
Thanks @Stranger6667 for submitting! Good catch. For now I will merge it but I believe that we should remove this line in requirements file because of #110.

PR #173, opened 2016-10-25
Change Customization topic example

lerrua:
Hi guys,

First, nice work with Djoser, this lib help us a lot!

So, I did this PR to update a Customization documentation to use an existing method because `send_email` method does not exist in RegistrationView.


haxoza:
We're sorry for long response time. Great catch! I'm merging it in!

Issue #147, opened 2016-05-26
Using Django Password Validators


pik:
I see Djoser has it's own method of providing password validation which does not use the Django `AUTH_PASSWORD_VALIDATORS` and will not accept identical arguments. Is there a reason for this? Wouldn't it be simpler to work directly with what is provided by Django auth (https://docs.djangoproject.com/en/1.9/topics/auth/passwords/) ?

mechtron:
+1

rawberg:
Looks like #107 was about to add this functionality but got combined with some other PRs and didn't end up getting merged unfortunately. I was able to follow the code in that commit and get it working pretty quickly by extending UserRegistrationSerializer and just overriding the password field as shown in the commit for #107.

aspcanada:
How do I set the validators that the rest of my app uses in AUTH_PASSWORD_VALIDATORS?

msolima:
You can do this in your settings.py

```
AUTH_PASSWORD_VALIDATORS = [
    {
        'NAME': 'django.contrib.auth.password_validation.MinimumLengthValidator',
    },
    {
        'NAME': 'django.contrib.auth.password_validation.CommonPasswordValidator',
    },
    {
        'NAME': 'django.contrib.auth.password_validation.NumericPasswordValidator',
    },
]
```

Then also in settings.py for DJOSER Settings:

```

from django.contrib.auth import password_validation

DJOSER = {
    'PASSWORD_RESET_CONFIRM_URL': '#/password/reset/confirm/{uid}/{token}',
    'ACTIVATION_URL': '#/activate/{uid}/{token}',
    'PASSWORD_VALIDATORS': [password_validation.validate_password]
}
```

works for me!

rvignesh89:
@msolima your fix worked for me as well. However when I wanted to add `UserAttributeSimilarityValidator` into the mix I realised that the fix was not that simple. It is because `UserAttributeSimilariryValidator` requires the User instance to be passed in. This does not happen currently since djoser uses serializer field validation which does not have access to user instance.

I like @pik's idea of re-using the `AUTH_PASSWORD_VALIDATORS ` settings in djoser considering DRY principles. Some digging into Django [documentation](https://docs.djangoproject.com/en/1.11/topics/auth/passwords/#integrating-validation) suggests us to use `validate_password` if we are writing our own mechanisms to change password which is exactly whats being done here.

Considering this I think we can implement password validation in djoser something like below and remove populating password validators in djoser settings again.


```
[...]

PR #176, opened 2016-11-10
Enable usage of Django Password Validators

agagata:
Ref #147

### DIFF
--- a/djoser/serializers.py
+++ b/djoser/serializers.py
@@ -28,7 +28,7 @@ class UserSerializer(serializers.ModelSerializer):
 class UserRegistrationSerializer(serializers.ModelSerializer):
     password = serializers.CharField(style={'input_type': 'password'},
                                      write_only=True,
-                                     validators=settings.get('PASSWORD_VALIDATORS'))
+                                     validators=utils.get_password_validators())
 
     class Meta:
         model = User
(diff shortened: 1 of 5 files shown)


mcgeeco:
So looks like we have two PR's for the same issue? This one looks better to me, for what it's worth.

djoser/serializers.py
```diff
@@ -28,7 +28,7 @@ class Meta:
 class UserRegistrationSerializer(serializers.ModelSerializer):
     password = serializers.CharField(style={'input_type': 'password'},
                                      write_only=True,
-                                     validators=settings.get('PASSWORD_VALIDATORS'))
+                                     validators=utils.get_password_validators())
```
piotr-szpetkowski: it would be awesome if you could change the hanging indents to be more consistent with rest of the project like:

```
password = serializers.CharField(
    style={'input_type': 'password'},
    write_only=True,
    validators=utils.get_password_validators()
)
```

djoser/serializers.py
```diff
@@ -124,7 +124,7 @@ def validate(self, attrs):
 
 class PasswordSerializer(serializers.Serializer):
     new_password = serializers.CharField(style={'input_type': 'password'},
-                                         validators=settings.get('PASSWORD_VALIDATORS'))
+                                         validators=utils.get_password_validators())
```
piotr-szpetkowski: Same thing as before - it would be great if you could change hanging indent.

piotr-szpetkowski:
Fixed with #205

PR #177, opened 2016-11-14
Enable usage of Django Password Validators

michalbon:
ref #147


msolima:
Could you please merge this request ? We need to be able to use Django password validators in djoser. 

Thanks

piotr-szpetkowski:
Fixed with #205

PR #189, opened 2017-04-28
Add support for plugable Token model.

edelvalle:


kowsari:
This would be very useful for me for switching over to use the django-rest-knox to generate different tokens for each device.

djoser/utils.py
```diff
@@ -21,13 +20,15 @@ def decode_uid(pk):
 
 
 def login_user(request, user):
-    token, _ = authtoken.models.Token.objects.get_or_create(user=user)
+    Token = import_string(settings.get('TOKEN_MODEL'))
```
piotr-szpetkowski: It would be awesome if you could remove duplicate definition of token (exactly same thing is happening at `logout_user`). My proposition for solution for this problem is to move it to the top of the file like:

```
from . import settings

Token = import_string(settings.get('TOKEN_MODEL'))
```
That way we will also avoid including another PEP-8 violation into djoser (local variable names should be all lowercase).

edelvalle: I was thinking the same, just waiting for someone to tell me how to do it... :blush:

edelvalle:
@piotr-szpetkowski Well I did some more changes, now `settings.get` is able to import objects if a `load` parameter is provided and you can ask for nested keys like `"SERIALIZERS.user"`.

In python 3.2 and up it does caching of the results, so it should work fast.

piotr-szpetkowski:
Hi. I've merged your pluggable token model via 10501c0d17a7b3035db7fe159cf5dd95ab042d46. Thank you very much for this contribution!

About dropping serializers manager - yes, I totally love the idea, however after discussing the matter internally and after recent settings refactor it seems that unfortunately we won't be able to merge your serializers code 😞 We're however working on a solution based on yours, so your hard work will not go to waste.

edelvalle:
@piotr-szpetkowski ahh no problem...! Thanks a lot; I'm glad I could help!

Issue #136, opened 2016-04-01
Clarification about REQUIRED_FIELDS


esistgut:
If I am not misunderstanding something In the default behaviour REQUIRED_FIELDS usage [should be limited to createsuperuser](https://docs.djangoproject.com/en/1.9/topics/auth/customizing/#django.contrib.auth.models.CustomUser.REQUIRED_FIELDS).
In djoser this seems to be used in non super user registrations too, how can I specify required fields for users registration without forcing them on super users too?

piotr-szpetkowski:
I believe that the solution which would cover most use cases would be to drop `REQUIRED_FIELDS` use and displace the responsibility into custom serializers provided via [SERIALIZERS setting](http://djoser.readthedocs.io/en/latest/settings.html#serializers).

piotr-szpetkowski:
As of the daa3c7de17642e605e5c519f12c06aad87cc1868 - after 2.0 is released you will need to override the serializer using inheritance and the [SERIALIZERS setting](http://djoser.readthedocs.io/en/latest/settings.html#serializers) mentioned above.

PR #200, opened 2017-06-20
Remove User.REQUIRED_FIELDS from fields attribute on user serializers

KaczuH:
User.REQUIRED_FIELDS is meant to be used only by createsuperuser management command

Resolves: #136


mcgeeco:
But won't this break if the User.EMAIL_FIELD isn't set?

dekoza:
@KaczuH can you update this for djoser-2.0 if it's still relevant? I'm closing this as it targets the legacy version.

Issue #155, opened 2016-06-20
django.conf.settings and djoser.settings


guillaumevincent:
django doc recommended not to import our  own settings file.

> Also note that your code should not import from either global_settings or your own settings file. django.conf.settings abstracts the concepts of default settings and site-specific settings; it presents a single interface. It also decouples the code that uses settings from the location of your settings.

https://docs.djangoproject.com/en/dev/topics/settings/#using-settings-in-python-code

We discuss this in the PR https://github.com/sunscrapers/djoser/pull/141

I look at the DRF [settings file](https://github.com/tomchristie/django-rest-framework/blob/bb56ca46ed6c07db0146dbdc61c672ff25f127de/rest_framework/settings.py) and I understand the objective with the `settings.get(...)` method in Djoser. It helps Djoser code to load user custom setting or the default one.

The implementation of DRF use an api_settings object with custom `__getattr__`

    from rest_framework.settings import api_settings
    print(api_settings.DEFAULT_RENDERER_CLASSES)

Djoser use a get method

    from djoser import settings
    print(settings.get('ACTIVATION_URL'))

The implementation is different, but it works the same way.

I will look if it is possible to use `django.conf.settings` but I have a doubt with the order loads of settings files.

PR #203, opened 2017-06-21
Kn/settings rewrite

KaczuH:
I saw a few problems with how we handle settings in Djoser. And decided to try solve some of them.

- `settings.get("SEND_ACTIVATION_EMAIL")` is cumbersome interface and it is used a lot across the code. We should look for more clear and simpler way. Thats why I decided to use `config.SEND_ACTIVATION_EMAIL` somewhat resembling how Django does.
- `merge_settings_dicts` was invoked every time an setting value is looked up. This is not optimal. Additionaly djoser settings aren't so complex so why to use this recursive function? I removed it and replaced with simple approach: instantiate default djoser settings and override anything that user customized in django settings.

There is some work to do but i want to show this idea for consideration.


mcgeeco:
For what it's worth, I'm definitely -1 on this. I think it's better to stick with the existing code which more closely resembles established Django patterns.

pawelswiecki:
Why `merge_settings_dicts` was there is the first place?

djoser/serializers.py
```diff
@@ -61,12 +62,9 @@ def create(self, validated_data):
...
     def perform_create(self, validated_data):
-        with transaction.atomic():
-            user = User.objects.create_user(**validated_data)
-            if settings.get('SEND_ACTIVATION_EMAIL'):
-                user.is_active = False
-                user.save(update_fields=['is_active'])
-        return user
+        if config.SEND_ACTIVATION_EMAIL:
+            validated_data.update({'is_active': False})
+        return User.objects.create_user(**validated_data)
```
haxoza: AFAIK we cannot get rid of `with transaction.atomic():` context manager here because `perform_create` method is called by `create` method where `IntegrityError` is handled. If we do not use separate transaction here then (depending on global django settings) the current transaction (e.g. per request) can be not valid anymore (e.g. for postgres).

djoser/serializers.py
```diff
@@ -62,9 +62,13 @@ def create(self, validated_data):
...
-        if config.SEND_ACTIVATION_EMAIL:
-            validated_data.update({'is_active': False})
-        return User.objects.create_user(**validated_data)
+        with transaction.atomic():
+            user = User.objects.create_user(**validated_data)
+            if config.SEND_ACTIVATION_EMAIL:
+                user.is_active = False
+                user.save(update_fields=['is_active'])
+        return user
+
```
haxoza: It could be almost like you did:

```
with transaction.atomic():
    if config.SEND_ACTIVATION_EMAIL:
        validated_data.update({'is_active': False})
    return User.objects.create_user(**validated_data)
```

haxoza:
@pawelswiecki good catch! `merge_settings_dicts` was there in order to allow djoser users easy default settings overriding. After current changes what user should do if she/he wants to change only one serializer?

KaczuH:
@haxoza Overriding settings is not changed from user perspective. Just define 

`DJOSER = {
    'SERIALIZERS': {
        'user': <user_custom_serializer>
}
}`

And only this one serializer is overriden default settings are intact.

djoser/settings.py
```diff
@@ -32,35 +38,49 @@
...
+                value.update(overriden_setting_value)
+            setattr(self, overriden_setting_name, value)
+
+
+class LazySettings(LazyObject):
+    def _setup(self, explicit_overriden_settings=None):
+        self._wrapped = Settings(default_settings, explicit_overriden_settings)
+
+
+config = LazySettings()
```
piotr-szpetkowski: `config` name seems very un-pythonic to me. Maybe something like `djoser_settings` would be better?

piotr-szpetkowski:
@mcgeeco care to elaborate what exactly is wrong with idea provided in this PR?

mcgeeco:
@piotr-szpetkowski No I see what the idea is now, it makes sense - I mainly thought it wasn't useful to go from `settings` to `config` as people might be used to looking for `settings` when they're overriding things.

haxoza:
See also #155

djoser/settings.py
haxoza: So, do we change `config` to `djoser_settings` or `settings`?

piotr-szpetkowski: settings sounds good in my opinion

djoser/settings.py
```diff
@@ -32,35 +38,50 @@
...
+
+
 def get(key):
-    user_settings = merge_settings_dicts(
-        deepcopy(default_settings), getattr(settings, 'DJOSER', {}))
+    """
+    This function is here only to provide backwards compatibility in case anyone uses old settings interface.
+    It is strongly encouraged to use dot notation.
+    """
+    warnings.warn('The settings.get(key) is superseded by the dot attribute access.')
```
piotr-szpetkowski: How about using [deprecation warning](https://docs.python.org/2/library/exceptions.html#exceptions.DeprecationWarning)?

djoser/settings.py
```diff
@@ -32,35 +38,50 @@
...
     try:
-        return user_settings[key]
-    except KeyError:
-        raise ImproperlyConfigured('Missing settings: DJOSER[\'{}\']'.format(key))
+        return getattr(config, key)
+    except ArithmeticError:
```
piotr-szpetkowski: Are you sure it's `ArithmeticError`? Doesn't make much sense tome.

djoser/settings.py
haxoza: also one more different idea: what about `djoser.conf.settings` like Django does `django.conf.settings`?

KaczuH: @haxoza @piotr-szpetkowski `djoser.conf.settings` would be the best but it will introduce incompatibility for older versions. 
I'll add deprecation warnings and try to keep old interface but it should be dropped in the future.

djoser/settings.py
KaczuH: @piotr-szpetkowski I'll add [PendingDeprecationWarning](https://docs.python.org/2/library/exceptions.html#exceptions.PendingDeprecationWarning)

Issue #186, opened 2017-02-19
Detect invalid reset token early, before asking the user to type new password


mnieber:
Trying to use a password reset token twice produces a 400 error. So far, so good.

However, I would like to detect this situation early, so I can prevent the situation where the user types their new password, only to find out that they need to request a new token (instead, I will tell them immediately).

Is that possible?

piotr-szpetkowski:
Hello. To answer your question - yes it is entirely possible to do that, however it would most probably involve creating new API endpoint, which would test if given token is correct or to update existing endpoint.

The problem is that in case if your user have not yet used the token he will call the endpoint twice (or two endpoints) and while such a simple check should not bring overwhelming load onto your infrastructure it still seems suboptimal.

Other solution which is more of a workaround might be to store used tokens at front-end, but here also are some pitfalls - e.g. user can remove cached data, limited to single device...

haxoza:
@piotr-szpetkowski In general I agree with your point of view. However with regards to:

> The problem is that in case if your user have not yet used the token he will call the endpoint twice (or two endpoints)

I think it is a matter of API client implementation. It can but it does not have to use this new additional endpoint.

mnieber:
One possibility would be add an optional `dry-run` boolean argument that allows you to simulate the reset, without actually resetting the password

piotr-szpetkowski:
Reference implementation provided with #204. Feature itself is not in scope of djoser core, however feel free to build an extension with what was provided in docs.

mnieber:
Sorry for the late reply, it looks good!

PR #204, opened 2017-06-22
Add examples chapter to docs for various edge cases users requests

KaczuH:
#186

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -10,7 +10,12 @@ to approach the problem like so:
 
 .. code-block:: python
 
+    from django.contrib.auth.tokens import default_token_generator
+    from rest_framework import generics, permissions, status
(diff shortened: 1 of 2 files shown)


docs/source/examples.rst
```diff
@@ -0,0 +1,26 @@
...
+
+Early detecting invalid password reset tokens
+---------------------------------------------
+
+When there is need to check if password reset token is still valid without actually resetting the password it is possible
+to approach the problem like so:
+
+.. code-block:: python
+
+    from djoser import serializers
```
piotr-szpetkowski: Since you include imports in your snippet I'd also add other required imports:

``` python
from django.contrib.auth.tokens import default_token_generator
from rest_framework import generics, permissions, status
from rest_framework.response import Response
```

Issue #185, opened 2017-01-31
switching from session to basic authentication makes application vulnerable to csrf 


amineparis:
how to use csrf built-in protection after switching to basic since this feature is included in seesion authentication

KaczuH:
Currently djoser supports only TokenAuthentication and it is possible to add JWT Authentication. See: [Djoser documentation](http://djoser.readthedocs.io/en/latest/authentication_backends.html)

Therefore your question is not quite valid in context of djoser usage.

Nevertheless we can discuss your problem:
    First of all:

1. Using of HTTP Basic Authentication is discouraged. See:

    -  [DRF documentation](http://www.django-rest-framework.org/api-guide/authentication/#basicauthentication)
    - [HTTP Basic Auth RFC](https://tools.ietf.org/html/rfc2617#section-3.1.1)
    - [OWASP](https://www.owasp.org/index.php/Basic_Authentication)

2. HTTP Basic authentication is not vulnerable to CSRF attack since it doesn't use cookies. You should use secure connection though. For further instructions see DRF and RFC documentation (Links above).


We hope that above information shed some light on your problem.

piotr-szpetkowski:
We've removed basic auth from supported authentication backends, therefore I'm closing this issue.

PR #214, opened 2017-07-11
Update docs to remove Basic Authentication

oliver-zhou:
As I saw with regards to @KaczuH comments in #185 - the docs should no longer mention Basic Authentication to avoid confusion, and instead mention using JWT as a possibility

### DIFF
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -17,8 +17,8 @@ Available endpoints
 Supported authentication backends
 ---------------------------------
 
-* `HTTP Basic Auth <http://www.django-rest-framework.org/api-guide/authentication/#basicauthentication>`_ (Default)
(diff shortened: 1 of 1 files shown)


piotr-szpetkowski:
I agree that Basic Auth should be dropped and we should not encourage using it. I'm not sure about listing JWT as supported since at the moment we do not provide out of the box support for it as mentioned in the docs: http://djoser.readthedocs.io/en/latest/authentication_backends.html#json-web-token-authentication

What do you think @sunscrapers/developers ?

piotr-szpetkowski:
I've merged your proposition via ecd4850d5c2881cef997a796520d8b34bc160e71 without mentioning JWT since we do not yet explicitly support it. I've also updated getting started docs page to include a discourage from using basic auth with 2d31e1caeca762b557ffe42b9974afa3d9254273.

Issue #213, opened 2017-07-11
/auth/register should have the option to not respond if username already exists


oliver-zhou:
If for example, I'm using email addresses as the Username, I wouldn't want the system to respond that "A user with that username already exists.". It's the same problem that PASSWORD_RESET_SHOW_EMAIL_NOT_FOUND is trying to solve

piotr-szpetkowski:
> I wouldn't want the system to respond that "A user with that username already exists."
Could you provide some background on the issue? Is it security or something else?

oliver-zhou:
For security : for those with Username == Email, we should give the developer the option to hide from the user that "Username is not found"

PR #215, opened 2017-07-12
Reregistration Flows

oliver-zhou:
Changes to address #213 

REREGISTRATION_SHOW_RESPONSE
----------------------------

If ``False`` (default), the ``/register/`` endpoint will always return
a ``HTTP_201_CREATED`` response, as well as the UserSerializer serializer.data response 
of one of those users, so that it makes it difficult to distinguish if a User has this email

**Default**: ``True``

SEND_REREGISTRATION_EMAIL
-------------------------

If ``True``, register endpoint will send warning and reminder email to user. 
Only active users are able to reset their passwords (User.is_active is True).
The email will ask the User to reset their password since someone is trying to
re-register their account.
If User.is_active is False, will send a warning

**Default**: ``False``

Reregistration Email is also different - if the User is_active or not, they will either have the option to reset password or will be reminded to complete activation.

### DIFF
--- a/djoser/conf.py
+++ b/djoser/conf.py
@@ -15,6 +15,7 @@ default_settings = {
     'USE_HTML_EMAIL_TEMPLATES': False,
     'SEND_ACTIVATION_EMAIL': False,
     'SEND_CONFIRMATION_EMAIL': False,
+    'SEND_REREGISTRATION_EMAIL': False,
     'SET_PASSWORD_RETYPE': False,
     'SET_USERNAME_RETYPE': False,
     'PASSWORD_RESET_CONFIRM_RETYPE': False,
@@ -38,6 +39,7 @@ default_settings = {
     },
     'LOGOUT_ON_PASSWORD_CHANGE': False,
     'USER_EMAIL_FIELD_NAME': 'email',
+    'REREGISTRATION_SHOW_RESPONSE': True,
 }
 
 SETTINGS_TO_IMPORT = ['TOKEN_MODEL']
--- a/djoser/templates/reregistration_inactive_email_body.html
+++ b/djoser/templates/reregistration_inactive_email_body.html
@@ -1,0 +1,14 @@
+{% load i18n %}{% autoescape off %}
+<html>
+<body>
+<p>
+{% blocktrans %}You're receiving this email because someone tried to reregister your user account at {{ site_name }}.{% endblocktrans %}</p>
+{% blocktrans %}Please find the previously sent activation email, or request a new one to properly set this account up{% endblocktrans %}
+
+<p>{% trans "Your username, in case you've forgotten:" %} <b>{{ user.get_username }}</b></p>
+
+<p>{% trans "Thanks for using our site!" %}</p>
+
+<p>{% blocktrans %}The {{ site_name }} team{% endblocktrans %}</p>
+</body></html>
+{% endautoescape %}
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -54,19 +54,20 @@ class RegistrationView(generics.CreateAPIView):
     permission_classes = (
         permissions.AllowAny,
     )
+    _users = None
 
     def create(self, request, *args, **kwargs):
-        if not settings.REREGISTRATION_SHOW_RESPONSE:
-            try:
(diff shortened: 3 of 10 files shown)


docs/source/settings.rst
```diff
@@ -56,6 +56,26 @@ If ``True``, register or activation endpoint will send confirmation email to use
...
 **Default**: ``False``
 
+SEND_REREGISTRATION_EMAIL
+-------------------------
+
+If ``True``, register endpoint will send warning and reminder email to user. 
+Only active users are able to reset their passwords (User.is_active is True).
+The email will ask the User to reset their password since someone is trying to
+re-register their account.
+If User.is_active is False, will send a warning
```
piotr-szpetkowski: Some of code values are decorated with higlight like \`\`True\`\`, but it seems that you have forgot about the others.

djoser/views.py
```diff
@@ -54,6 +54,22 @@ class RegistrationView(generics.CreateAPIView):
...
+            email_users = self.get_email_users(request.data.get('email'))
+            for user in email_users:
+                serializer = self.get_serializer(instance=user)
+                if settings.SEND_REREGISTRATION_EMAIL:
+                    self.send_reregistration_email(user)
+                headers = self.get_success_headers(serializer.data)
+            if not settings.REREGISTRATION_SHOW_RESPONSE and email_users:
+                return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+        except User.DoesNotExist:
+            pass
```
piotr-szpetkowski: I'm wondering if this `try ... except ...` block could not be replaced with some smart `if` statement, since you are not doing anything with the exception.

djoser/templates/reregistration_inactive_email_body.html
```diff
@@ -0,0 +1,14 @@
...
+{% blocktrans %}You're receiving this email because someone tried to reregister your user account at {{ site_name }}.{% endblocktrans %}</p>
+{% blocktrans %}Please find the previously sent activation email, or request a new one to properly set this account up{% endblocktrans %}
+
+<p>{% trans "Your username, in case you've forgotten:" %} <b>{{ user.get_username }}</b></p>
+
+<p>{% trans "Thanks for using our site!" %}</p>
+
+<p>{% blocktrans %}The {{ site_name }} team{% endblocktrans %}</p>
+</body></html>
+{% endautoescape %}
```
piotr-szpetkowski: Also, wouldn't we prefer to resend activation email in case of such API call? Emails tend to (rarely) get lost e.g. temporary mail server downtime.

It might be a good idea to have a discussion on that @haxoza @KaczuH

djoser/conf.py
```diff
@@ -15,6 +15,7 @@
     'USE_HTML_EMAIL_TEMPLATES': False,
     'SEND_ACTIVATION_EMAIL': False,
     'SEND_CONFIRMATION_EMAIL': False,
+    'SEND_REREGISTRATION_EMAIL': False,
```
oliver-zhou: Sounds fine to me - will try to keep nouns consistent and modify the verbs if it pops up elsewhere as well.

djoser/conf.py
```diff
@@ -38,6 +39,7 @@
     },
     'LOGOUT_ON_PASSWORD_CHANGE': False,
     'USER_EMAIL_FIELD_NAME': 'email',
+    'REREGISTRATION_SHOW_RESPONSE': True,
```
oliver-zhou: I had it mixed up in what I wrote for the docs : the **Default** was set to True, but in-line the description was False. Fixing the docs to match.

oliver-zhou:
I have some time to revisit this pull request, and bring this up to speed with Djoser 1.15  - before I spend the investment - what are the plans for the branches in development right now? If I'm working off 1.15, will it be future resistant - or should I wait for 2.0?

zefciu:
@oliver-zhou It would be great to pull this into 1.0 branch. The 2.0 needs to be refactored so there is no point in waiting with it.

dekoza:
@oliver-zhou Sorry for such long delay with this PR. Do you feel it is suitable for version 2.0 that was released recently? I'd like to revisit this feature request.

PR #224, opened 2017-08-13
Simplified del to pop

davidtgq:
Didn't test, but should work because self.fields returns a dictionary.

Also if you read this, is there any documentation or explanation why it's necessary to rename the username key to new_username?


piotr-szpetkowski:
Hi. There isn't any documentation on that, however as far as I know the original idea was that the view expects a field called `new_username` to avoid a situation, where the username would simply be renamed within the serializer.

This is going to be replaced in future with better and more organized structure, however for now, until that, we'll have to keep it as it is.

It seems like patch is good to merge, so I'll probably merge it this evening - thank you for the contribution!

Issue #223, opened 2017-08-04
PasswordResetView proposal to move the User.is_active check from database to Python


davidtgq:
Something like this:

```
class PasswordResetView(utils.ActionViewMixin, generics.GenericAPIView):
    def get_users(self, email):
        if self._users is None:
            active_users = User._default_manager.filter(
                email__iexact=email,
                # is_active=True, <--- v0.6 code
            )
            # self._users = [u for u in active_users if u.has_usable_password()] <--- v0.6 code
            self._users = [u for u in active_users if u.is_active and u.has_usable_password()]
        return self._users
```

Here are the reasons:

1. If `is_active` is not True, then this opportunity can be used to notify the user that they've already registered an account with that email address, and [offer to] resend the activation email.

```
_users = User._default_manager.filter(email=email)
self._users = []
for u in _users:
    if u.has_usable_password():
        if u.is_active:
            self._users.append(u)
        else:
            # TODO resend activation email
```

2. I think with the v0.6 code, there will be some conflict if `is_active` is implemented in a model like this:

```
class User(AbstractBaseUser):
    score = models.SmallIntegerField(default=0)
    status = models.SmallIntegerField(default=0)
    email = models.CIEmailField(unique=True)
    name = models.CITextField()
    ip = models.ArrayField(models.GenericIPAddressField(unpack_ipv4=True))

    objects = UserManager()

    USERNAME_FIELD = 'email'

    @property
    def is_active(self):
        # FIXME can't do this, djoser uses .filter(is_active=True)
        return self.status > 0

    @is_active.setter
    def is_active(self, value):
        if value is True:
            self.status = abs(self.status or 1)
        elif value is False:
            self.status = -abs(self.status or 0)
        else:
            self.status = value
```
3. Any concerns about performance/inefficiency of retrieving unnecessary rows from the database should be nominal in real-world usage because there wouldn't be very many unactive accounts tied to a single email address in most cases. And again, if the activation email resend is implemented, then the retrieval of these rows wouldn't be considered wasted.

piotr-szpetkowski:
Hi. At the first glance your proposition seems very interesting and I thought that we might go along with something like this. However - after taking a deeper look into the djoser codebase I'm afraid that there is a bigger problem in djoser codebase.

[...]

PR #225, opened 2017-08-18
Move User.is_active check from database to Python

davidtgq:
https://github.com/sunscrapers/djoser/issues/223

Also renamed `active_users_kwargs` to `filter_users_kwargs`, and `active_users` to `filtered_users`, because the "active" word is no longer applicable. Feel free to change these variable names to something else if necessary.

### DIFF
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -130,11 +130,11 @@ class PasswordResetView(utils.ActionViewMixin, generics.GenericAPIView):
     def get_users(self, email):
         if self._users is None:
             email_field_name = get_user_email_field_name(User)
-            active_users_kwargs = {
-                email_field_name + '__iexact': email, 'is_active': True
+            filter_users_kwargs = {
+                email_field_name + '__iexact': email
             }
-            active_users = User._default_manager.filter(**active_users_kwargs)
-            self._users = [u for u in active_users if u.has_usable_password()]
+            filtered_users = User._default_manager.filter(**filter_users_kwargs)
+            self._users = [u for u in filtered_users if u.is_active and u.has_usable_password()]
         return self._users
 
     def send_password_reset_email(self, user):
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -130,12 +130,12 @@ class PasswordResetView(utils.ActionViewMixin, generics.GenericAPIView):
     def get_users(self, email):
         if self._users is None:
             email_field_name = get_user_email_field_name(User)
-            filter_users_kwargs = {
+            users = User._default_manager.filter(**{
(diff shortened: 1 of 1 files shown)


djoser/views.py
```diff
@@ -130,11 +130,12 @@ def _action(self, serializer):
...
         if self._users is None:
             email_field_name = get_user_email_field_name(User)
-            active_users_kwargs = {
-                email_field_name + '__iexact': email, 'is_active': True
+            filter_users_kwargs = {
+                email_field_name + '__iexact': email
             }
-            active_users = User._default_manager.filter(**active_users_kwargs)
-            self._users = [u for u in active_users if u.has_usable_password()]
+            _users = User._default_manager.filter(**filter_users_kwargs)
```
piotr-szpetkowski: There is no need to start local variable name with underscore. In my opinion it is only meaningful in class/instance scope and indicates if a member should be treated like a private.

djoser/views.py
```diff
@@ -130,11 +130,12 @@ def _action(self, serializer):
     def get_users(self, email):
         if self._users is None:
             email_field_name = get_user_email_field_name(User)
-            active_users_kwargs = {
-                email_field_name + '__iexact': email, 'is_active': True
+            filter_users_kwargs = {
+                email_field_name + '__iexact': email
             }
```
piotr-szpetkowski: Since these kwargs are rather short now, you might as well simplify the `users` into:
```
users = User._default_manager.filter(**{email_field_name + '__iexact': email})
```
with a line break if necessary.

That way, it's still readable with less lines to read.

djoser/views.py
```diff
@@ -130,11 +130,12 @@ def _action(self, serializer):
...
-            active_users = User._default_manager.filter(**active_users_kwargs)
-            self._users = [u for u in active_users if u.has_usable_password()]
+            _users = User._default_manager.filter(**filter_users_kwargs)
+            self._users = [u for u in _users
+                           if u.is_active and u.has_usable_password()]
```
piotr-szpetkowski:
```
self._users = [
    u for u in users if u.is_active and u.has_usable_password()
]
```

would seem a little bit cleaner to me, but I'll leave it up to your taste.

djoser/views.py
davidtgq: Yup this needed a line break, unless `email_field_name` is renamed to something <= 5 characters to fit in one line

davidtgq:
Thanks for the feedback, I've made the changes

piotr-szpetkowski:
Awesome! Everything is great, so I'm merging your PR into master. Again, thank you very much for the contribution.

PR #236, opened 2017-10-18
Update default serializers in docs

ir4y:


piotr-szpetkowski:
Hi @ir4y.
Thank you for taking care of this - somehow I missed updating these docs.

PR #241, opened 2017-10-25
Update docs for new URLConf in Djoser 1.0.0.

### LINKED ISSUE #240: Proper way to upgrade from 0.7.0 to 1.0.1?
Hi - thanks for building Djoser - it's a great little library! It would be helpful in the Changelog to show breaking changes, or to show a note or two on upgrading from version to version. I used to use the various `authtoken` URLs in my app, but now it appears there is only one available - `/auth/token/create`. 

Is there a new way of performing actions like password reset and creating a new user? `/auth/token/create` only appears to replace the old `/auth/login` URL. 

Happy to add this to the changelog if I could get a little guidance on how to update for the latest version :).

YPCrumble:
@piotr-szpetkowski adding these lines would have probably solved my issue. Feel free to pull in or not as you see fit. If you'd suggest any changes please let me know as well.

Thanks again for building this great library!

Fixes #240

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ This project adheres to `Semantic Versioning <http://semver.org/>`_.
 `1.0.0`_ (2017-10-14)
 ---------------------
 
+* **Breaking**: For Token-based and JWT authentication ``djoser.urls`` should be included in the URLconf as well as either ``djoser.urls.authtoken`` or ``djoser.urls.jwt``
 * Add JWT authentication support
 * Add/Update documentation about JWT
 * Add/Update/Fix tests where necessary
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -24,7 +24,7 @@ This project adheres to `Semantic Versioning <http://semver.org/>`_.
 * Replace built-in email support with django-templated-mail
 * Refactor test configuration for better performance and organization
 * Refactor RootView to have better support for modular URLs
-* Update URLs to be slightly more RESTful.
+* Update URLs to be slightly more RESTful. **NOTE**: For Token-based and JWT authentication, this means that 'djoser.urls' should be included in the URLconf as well as either 'djoser.urls.authtoken' or 'djoser.urls.jwt'
```
piotr-szpetkowski: Just to make it more readable and visible - could we perhaps move it to the top and slightly reformat it to something alike:

```
* **Breaking**: For Token-based and JWT authentication, this means that ``djoser.urls`` should be included in the URLconf as well as either ``djoser.urls.authtoken`` or ``djoser.urls.jwt``
```

Besides that, it's awesome and I'd love to merge it.

YPCrumble:
Thanks @piotr-szpetkowski - just updated.

PR #242, opened 2017-11-01
Refactored Email into Mixin

hammadzz:
Refactored email into a mixin so it can easily be overridden. Currently a lot of methods in views need to be overridden just to replace the email template and such. This abstracts out the email functionality to be easily modifiable.

Subclass the view and override `get_email(self):` with your own email.py


piotr-szpetkowski:
Resolved in #243

PR #246, opened 2017-11-03
Raise AuthenticationFailed instead of ValidationError.

annaelde:
Currently, failure to authenticate a username/password to issue a token results in a 400 Bad Request status code. I think that a 401 Unauthorized response makes more sense in this context. To return the desired status code, I overrode ````self.fail```` to raise an AuthenticationFailed exception.

### DIFF
--- a/djoser/serializers.py
+++ b/djoser/serializers.py
@@ -109,6 +109,12 @@ class TokenCreateSerializer(serializers.Serializer):
     def _validate_user_is_active(self, user):
         if not user.is_active:
             self.fail('inactive_account')
+    
+    def fail(self, key, **kwargs):
+        try:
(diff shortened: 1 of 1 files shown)


piotr-szpetkowski:
Interesting idea, however unfortunately it's not something that we can consider at the moment. Since 1.0 release we should keep backward compatibility both feature-wise and API-wise and in case of changing status code we'd force Djoser API consumers to update. It is not desirable, because with mentioned 1.0 release we've made an implicit promise that API is stable.

If you believe that this should be discussed further it would be great if you would open a new issue.

Thank you though for spending effort to work on this - sometime in future near 2.0 release we might get back to this and consider it again.

annaelde:
I understand, thank you for the explanation. Would be glad to see this in 2.0 in the future 😄

PR #250, opened 2017-11-20
Use HTTP Origin header for password reset link domain

hbielenia:
This commit changes domain URL source in password reset e-mail, so that it points to domain received in HTTP Origin header.

It attempts to solve the problem of sending e-mail with frontend domain name, not the backend domain that djoser runs under, much like #158 but with simpler settings and no merge conflicts. We rely on client sending a correct HTTP Origin header, which exists exactly for this purpose. The header is documented and supported by all major browsers: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin

I also added a `PASSWORD_RESET_DOMAINS` setting to djoser config that is a list containing allowed domains, to avoid opening a hole that would allow abusing this feature for phishing attacks. If domain isn't in the list, it's ignored and previous method of determining domain is used.


piotr-szpetkowski:
Hi @hbielenia.
Thanks for picking it up. It's a great solution and I'd love to have it in djoser. I just wonder if we couldn't make it a bit more generic, since activation email also makes use of frontend domain.

PR #254, opened 2017-12-17
make perform_create use the same model name in Meta

Alwerdani:
If you have multiple user types need to custom `UserCreateSerializer` for being used those types,  you need to change model name, so you must also override `perform_create` method cuz the model name is hardcoded in this method

This PR is fixing this and get the model name from the class Meta


Alwerdani:
No need do that, cuz I can override the save method in the base Class and create the sub Classes

PR #255, opened 2017-12-30
Fix a typo error

taoufik07:
I noticed a typo error in the docs and I think it's better now !


piotr-szpetkowski:
Thank you, nice catch! 👍

PR #266, opened 2018-02-15
Fix validation: keep original ValidationError

blueyed:
The casting into serializers.ValidationError was losing the code(s) etc,
and Django's ValidationError should get handled correctly by DRF.

TODO:
 - [ ] (adjust?) test(s)

### DIFF
--- a/djoser/serializers.py
+++ b/djoser/serializers.py
@@ -47,16 +47,9 @@ class UserCreateSerializer(serializers.ModelSerializer):
             User.USERNAME_FIELD, User._meta.pk.name, 'password',
         )
 
-    def validate(self, attrs):
-        user = User(**attrs)
-        password = attrs.get('password')
-
-        try:
-            validate_password(password, user)
-        except django_exceptions.ValidationError as e:
-            raise serializers.ValidationError({'password': list(e.messages)})
-
-        return attrs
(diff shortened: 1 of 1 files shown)


zefciu:
@blueyed In fact, the code of the error is lost much earlier. The `validate_password` method loses it when it combines various errors into one. I will try to research a method to preserve this code, however, as there should be some way according to django design.

blueyed:
@zefciu 
Are you sure?
AFAIR `validate_password` from Django will have it, and then it needs to be handled properly further down.  E.g. https://github.com/encode/django-rest-framework/pull/5785 is necessary IIRC.

blueyed:
Django code:
https://github.com/django/django/blob/36fceeec883c5082168714a0eb14a2fe40f9d79b/django/contrib/auth/password_validation.py#L35-L51

zefciu:
From what I researched when the errors are combined in line 51 the code is no longer there. Tomorrow I will look at the logic of creating one ValidationError from a list to understand what is and what should be happening here.

zefciu:
There is a canonical way of error conversion that is implemented in django-rest-framework. I used it in this PR: https://github.com/sunscrapers/djoser/pull/337 @blueyed please confirm that this fixes your problems.

PR #280, opened 2018-05-22
Change customization terminology for Djoser 1.0+

oliver-zhou:
Change from Registration to UserCreate


dekoza:
Can't accept it as it breaks the API. If someone subclassed the view or uses the URL, they will suddenly have problems.

PR #281, opened 2018-05-25
catch exception where TOKEN_MODEL not in use for UserDeleteView

garyburgmann:
I am using rest_framework_jwt instead, this works fine for me


zefciu:
Catching *every possible exception* is certainly not a good idea. Also, please elaborate on the use-case.

PR #286, opened 2018-07-18
Add django-2.0 to tox matrix, update requirements, bump version

dekoza:


testproject/settings.py
```diff
@@ -19,10 +22,11 @@
 
 SECRET_KEY = '_'
 
-MIDDLEWARE = [
+MIDDLEWARE_CLASSES = MIDDLEWARE = [
```
haxoza: Do we really need `MIDDLEWARE_CLASSES`?

testproject/settings.py
```diff
@@ -1,6 +1,9 @@
 from __future__ import absolute_import
 import os
+import sys
 
+sys.path.append('../')
+sys.path.append('./')
```
haxoza: Why?

README.rst
```diff
@@ -35,8 +35,8 @@ Requirements
 To be able to run **djoser** you have to meet following requirements:
 
 - Python (2.7, 3.4, 3.5, 3.6)
-- Django (1.10, 1.11)
-- Django REST Framework (3.7)
+- Django (1.11, 2.0)
```
haxoza: 1.10 is removed from supported versions in README but I can see it in tox dependencies.

.gitignore
```diff
@@ -1,40 +1,197 @@
...
+/site
+
+# mypy
+.mypy_cache/
+
+### Python Patch ###
+.venv/
+
+
+# End of https://www.gitignore.io/api/django,python,pycharm+all
```
haxoza: I'm not the fan of auto generated ignore file for libraries like that. I believe we will never ever use e.g. crashlitics in this project 😄

dekoza: I'm more for standards but I admit the .gitignore here might be overkill :wink:

dekoza:
Actually, I managed to update urls to more RESTful fashion, so this PR is no longer needed. Changes will be included in next PR with said urls.

PR #292, opened 2018-08-08
Update example with working password

superhawk610:
Using the username/password provided, djoser throws the following error:

```
{"password":["The password is too similar to the username.","This password is too short. It must contain at least 8 characters."]}
```

The first logical change is to just use `password` for the password, but that errors with

```
{"password":["This password is too common."]}
```

The change I've proposed (`alpine12`) works with no issues.

**Note** - I'm not sure which branch to PR, project maintainers feel free to change the base branch or request that I do so.

### DIFF
--- a/docs/source/sample_usage.rst
+++ b/docs/source/sample_usage.rst
@@ -28,7 +28,7 @@ Register a new user:
 
 .. code-block:: text
 
-    $ curl -X POST http://127.0.0.1:8088/auth/users/ --data 'username=djoser&password=djoser'
+    $ curl -X POST http://127.0.0.1:8088/auth/users/ --data 'username=djoser&password=alpine12'
     {"email": "", "username": "djoser", "id":1}
 
 So far, so good. We have just created a new user using REST API.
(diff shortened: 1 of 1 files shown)


superhawk610:
It seems like this is due to the `AUTH_PASSWORD_VALIDATORS` block included in the `settings.py` created by `django-admin startproject`, but I still think it's worth changing just to avoid potential errors.

---

```
curl -X GET http://localhost:8000/auth/users/me/
```

also initially fails since it returns a 301 redirect, so it probably wouldn't hurt to update that example to include the `-L` flag for curl to follow redirects.

```
curl -LX GET http://localhost:8000/auth/users/me/
```

dekoza:
You are right. The examples are from before `AUTH_PASSWORD_VALIDATORS` were introduced and have not been updated since. Thanks.

dekoza:
Pleas, add that "L" to your PR and I'll merge it.

superhawk610:
Done. I also updated the login example to use the new password.

Also wanna double-check that `1.x` is the correct branch to merge into.

dekoza:
`version 1.x` is OK. Thank you for your time.

PR #295, opened 2018-09-01
Add a test to make sure that DRF docs work

akalipetis:
When the DRF docs try to inspect the UserViewSet, get_serializer_class
crashes because there's no request present. In such case, we should
fallback to the default serializer and everything should work as expected.

Signed-off-by: Antonis Kalipetis <akalipetis@gmail.com>

### DIFF
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -285,14 +285,14 @@ class UserViewSet(UserCreateView, viewsets.ModelViewSet):
     def get_serializer_class(self):
         if self.action == 'create':
             return settings.SERIALIZERS.user_create
-        if self.action == 'remove' or (
+        elif self.action == 'remove' or (
                 self.action == 'me' and self.request and
                 self.request.method == 'DELETE'
         ):
             return settings.SERIALIZERS.user_delete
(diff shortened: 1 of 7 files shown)


akalipetis:
Not sure which is the best way to install dependencies, while docs say to use `pipenv`, Travis is using `requirements.txt`.

I have added `coreapi` to both `Pipfile` and `requirements.txt` and also locked `Pipfile.lock` again. Let me know if you'd like any changes in the code or setup 😀

djoser/views.py
```diff
@@ -285,12 +285,14 @@ def get_permissions(self):
     def get_serializer_class(self):
         if self.action == 'create':
             return settings.SERIALIZERS.user_create
-        elif self.action == 'remove' or (
-                self.action == 'me' and self.request.method == 'DELETE'):
+        if self.action == 'remove' or (
```
dekoza: what's wrong with `elif`?

dekoza:
Thanks for your time :)

akalipetis:
Thanks a lot for a great library!

PR #297, opened 2018-09-05
Split user serializers (fix #294)

### LINKED ISSUE #294: Split user serializers
In many applications, the response from `/users/<id>/` should contain a different set of fields than the response from `/users/me/`. This could be for privacy (not allowing other users access to confidential attributes on the user model) or simply because one requires a different set of information for the current user (e.g. settings).

It is possible to work around this issue with a custom viewset, but it does not work on legacy endpoints and adds unnecessary complexity to this common use case.

I propose that the serializer configuration for users be split into the following two settings:
- `user`
- `me`

joshua-s:
This adds a new option `current_user` to `settings.SERIALIZERS` that allows different information to be returned from 'me' endpoints.

Currently a work in progress. See #294.

### DIFF
--- a/djoser/conf.py
+++ b/djoser/conf.py
@@ -55,6 +55,8 @@ default_settings = {
             'djoser.serializers.UserDeleteSerializer',
         'user':
             'djoser.serializers.UserSerializer',
+        'current_user':
+            'djoser.serializers.CurrentUserSerializer',
         'token':
             'djoser.serializers.TokenSerializer',
         'token_create':
--- a/djoser/serializers.py
+++ b/djoser/serializers.py
@@ -1,3 +1,5 @@
+import warnings
+
 from django.contrib.auth import authenticate, get_user_model
 from django.contrib.auth.password_validation import validate_password
 from django.core import exceptions as django_exceptions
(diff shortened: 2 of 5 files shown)


joshua-s:
@dekoza I have a backwards compatibility dilemma here. It appears that `settings.SERIALIZERS.current_user` should be defaulted to `'djoser.serializers.CurrentUserSerializer'` like every other serializer setting. However, this will break backwards compatibility with users that have specified a custom `settings.SERIALIZERS.user` (they will suddenly start to receive different results from current user endpoints).

I have left the legacy `/me` untouched. It will still use `settings.SERIALIZERS.user`.

dekoza:
OK, we need a deprecation route here. For now, `settings.SERIALIZERS.current_user` should point to a proxy class that:

1. raises DeprecationWarning and points to documentation where the separation and rationale behind it should be described. 
2. routes all calls to the same class that is set in `settings.SERIALIZERS.user`

This way we will have a way to inform about the separation of serializers without breaking compatibility with custom settings. Can you do this or should I take over?

joshua-s:
@dekoza I have addressed your comments!

dekoza:
Awesome work! Thank you

Issue #299, opened 2018-09-07
User viewset does not support partial PATCH updates 


joshua-s:
## Code at fault:
https://github.com/sunscrapers/djoser/blob/ab825cb57785e77c8618abaf4301a25b66b4071f/djoser/views.py#L332-L333

## Expected result:
`PATCH` requests should support partial updates (where only included fields are validated). This allows one to update, say, the `first_name` field without needing to also include/update the `username`.

Supporting partial updates over `PATCH` is standard for Django Rest Framework. See: 
- https://github.com/encode/django-rest-framework/blob/eb3d078a2cbda2d6e576cbfd773f153d1bdf5c5b/rest_framework/routers.py#L122
- https://github.com/encode/django-rest-framework/blob/eb3d078a2cbda2d6e576cbfd773f153d1bdf5c5b/rest_framework/mixins.py#L82-L84

dekoza:
True, this needs to be updated. Thanks.

joshua-s:
@dekoza I sent a PR: #300

dekoza:
Good job, thanks!

PR #300, opened 2018-09-07
Use partial_update for PATCH on /me

joshua-s:
Fixes issue #299


joshua-s:
@dekoza would it be possible to release this as a quick update?

PR #305, opened 2018-09-14
Use django-rest-framework-simplejwt instead of django-rest-framework-jwt

### LINKED ISSUE #302: Use different JWT package
For all intents and purposes, django-rest-framework-jwt has been abandoned and is no longer receiving updates (https://github.com/GetBlimp/django-rest-framework-jwt/issues/449). It is a good time for djoser to consider another package for JWT authentication.

One that I found:
- https://github.com/davesque/django-rest-framework-simplejwt

joshua-s:
Fix #302

### DIFF
--- a/djoser/social/token/jwt.py
+++ b/djoser/social/token/jwt.py
@@ -1,9 +1,11 @@
 class TokenStrategy:
     @classmethod
     def obtain(cls, user):
-        from rest_framework_jwt.settings import api_settings
-        payload = api_settings.JWT_PAYLOAD_HANDLER(user)
+        from rest_framework_simplejwt.tokens import RefreshToken
+        from django.utils.six import text_type
+        refresh = RefreshToken.for_user(user)
         return {
-            'token': api_settings.JWT_ENCODE_HANDLER(payload),
-            'user': user
+            'access': text_type(refresh.access_token),
+            'refresh': text_type(refresh),
+            'user': user,
         }
--- a/Pipfile
+++ b/Pipfile
@@ -18,8 +18,8 @@ ipdb = "*"
 
 [packages]
 djangorestframework = "*"
-"djangorestframework-jwt" = "*"
 django = "*"
 "django-templated-mail" = "*"
 "social-auth-app-django" = "*"
 coreapi = "*"
+djangorestframework-simplejwt = "*"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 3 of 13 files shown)


joshua-s:
Currently, this is not backwards compatible because django-rest-framework-simplejwt returns a pair of tokens, `access` and `refresh`. The `refresh` token is the only one that can be used with the refresh endpoint.

djoser/social/token/jwt.py
```diff
@@ -1,9 +1,11 @@
 class TokenStrategy:
     @classmethod
     def obtain(cls, user):
-        from rest_framework_jwt.settings import api_settings
-        payload = api_settings.JWT_PAYLOAD_HANDLER(user)
+        from rest_framework_simplejwt.tokens import RefreshToken
```
dekoza: Can this be changed so that it stays backward-compatible? For example someone may use `JWT_PAYLOAD_HANDLER` - what then? If not, a deprecation route must be taken.

djoser/social/token/jwt.py
```diff
@@ -1,9 +1,11 @@
...
+        from django.utils.six import text_type
+        refresh = RefreshToken.for_user(user)
         return {
-            'token': api_settings.JWT_ENCODE_HANDLER(payload),
-            'user': user
+            'access': text_type(refresh.access_token),
```
dekoza: This breaks all apps that depend on `token` field. We need to keep it. Such changes should be made either using deprecation route or, if radical change is inevitable, a new major version needs to be released - we promised to use semantic versioning.

joshua-s: So, this JWT package, which appears to be the only actively maintained one for Django Rest Framework, splits tokens into `access` and `refresh`. Only the `refresh` tokens can be used at the `refresh` endpoint.

I think renaming `access` to `token` would still not be backwards compatible for this reason. Although, token refreshing is broken in django-rest-framework-jwt, so it might not be a big deal.

Ultimately, we need to decide if this is worth shipping a major version. Or, maybe we should take the hint that the community is moving to something else (no actively-maintained JWT packages) and just get rid of JWT completely in favor of a different system.

ozeranskiy:
I think using only django-rest-framework-simplejwt is a good idea.

joshua-s:
I resolved all conflicts so this is ready if it gets approved.

alexandernst:
@dekoza Can this get a new review? Our company is very interested in this feature and we're willing to actively help to push this to the end by throwing into it as many developers as needed.

ozeranskiy:
i think need make new major version with used django-rest-framework-simplejwt

dekoza:
This is a major change and probably breaking. I'll merge this into master now and consider bumping major version (and releasing it) during the weekend. I'd like to introduce some more breaking changes into 2.0 while we're at it :wink:

PR #307, opened 2018-09-19
views: use @action(detail=False) instead of deprecated @list_route

blueyed:
Fixes:

> …/djoser/views.py:336: PendingDeprecationWarning:
> `list_route` is pending deprecation and will be removed in 3.10 in
> favor of `action`, which accepts a `detail` bool. Use `@action(detail=False)`
> instead.

Have not checked if it works with all supported DRF versions, but CI should tell us.

### DIFF
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -3,7 +3,7 @@ from django.contrib.auth.tokens import default_token_generator
 from django.urls.exceptions import NoReverseMatch
 from django.utils.timezone import now
 from rest_framework import generics, permissions, status, views, viewsets
-from rest_framework.decorators import list_route
+from rest_framework.decorators import action
 from rest_framework.response import Response
 from rest_framework.reverse import reverse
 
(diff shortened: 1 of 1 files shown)


dekoza:
It doesn't work with DRF 3.7. Since DRF has no LTS version we try to keep compatibility with both current and previous minor version. When DRF 3.9 shows up, we'll drop support for DRF 3.7 and this PR will stop failing :wink:

dekoza:
It's not - mainly because [DRF takes care of handling `@list_route` and `@detail_route` properly ](https://github.com/encode/django-rest-framework/blob/master/rest_framework/decorators.py#L224)

blueyed:
Yeah.. it was meant to avoid the pendingdeprecationwarning, but I assume that does not make sense given that it is pending..?!

jaap3:
3.9 has been released, so maybe reopen this?

PR #326, opened 2018-12-04
fix logout Anonymous user

ckkz-it:


ckkz-it:
screwed up in last commit logout a bit, when sending signal, because user is Anonymous at the moment signal being sent
fixed now

PR #350, opened 2019-02-07
Fix broken current_user setting

### LINKED ISSUE #349: settings.SERIALIZERS.current_user is broken
In `UserViewSet`, `get_serializer_class` is meant to return `settings.SERIALIZERS.user` for actions involving other users or `settings.SERIALIZERS.current_user` for actions involving the current user.

However, this doesn't actually happen. This is the current code:
```
def get_serializer_class(self):
    if self.action == 'me':
        # Use the current user serializer on 'me' endpoints
        self.serializer_class = settings.SERIALIZERS.current_user

    if self.action == 'create':
        return settings.SERIALIZERS.user_create
```

Instead of returning the correct serializer class, it gets set manually with `self.serializer_class = ...`. This doesn't actually do anything because it gets overwritten by the default serializer class returned at the end of the method.

brortao:
Actually use the `current_user` serializer from settings when it is set.
Fixes #349.

### DIFF
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -289,10 +289,6 @@ class UserViewSet(UserCreateView, viewsets.ModelViewSet):
         return super(UserViewSet, self).get_permissions()
 
     def get_serializer_class(self):
-        if self.action == 'me':
-            # Use the current user serializer on 'me' endpoints
-            self.serializer_class = settings.SERIALIZERS.current_user
-
         if self.action == 'create':
             return settings.SERIALIZERS.user_create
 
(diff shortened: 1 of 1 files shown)


dekoza:
Please write a test case that exposes the bug but passes after your fix.

joshua-s:
@dekoza I am willing to write a test case for this if you could merge it

joshua-s:
I added a potential test in #354.

One question I had when writing it. It does not appear that any of the `me` endpoints are actually using the wrong serializer. Consequently, it is not failing on the current codebase. However, I do agree that this PR sets the serializer in a more proper way, so it should still be merged in my opinion.

Am I missing anything @brortao? How can I reproduce the error you are trying to fix?

brortao:
Thanks for writing the test, @joshua-s ! Apologies for my radio silence on the issue -- I'm no longer using Django or Djoser, so it took a bit to find time to look more into this.

I can't say why the test you wrote passes. Mostly to verify my own sanity, I made a minimal Django project that demonstrates the issue: https://github.com/brortao/djoser-bug

The bug does indeed still appear. I can't explain why it doesn't in the test. This demo project uses token-based authentication, if that changes anything.

joshua-s:
@brortao thanks for that project. I will check it out

PR #375, opened 2019-05-08
Replace deprecated `list_route` with `action(detail=False)` decorator.

MicahLyle:
DRF gives this warning:
```
RemovedInDRF310Warning: `list_route` is deprecated and will be removed in 3.10 in favor of `action`, which accepts a `detail` bool. Use 
`@action(detail=False)` instead.
    @list_route(['get', 'put', 'patch', 'delete'])
```

This fixes/removes those warnings by using the newer `action` decorator instead of `list_route`.

### DIFF
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -9,7 +9,7 @@ from rest_framework import (
     status, views,
     viewsets,
 )
-from rest_framework.decorators import list_route
+from rest_framework.decorators import action
 from rest_framework.response import Response
 from rest_framework.reverse import reverse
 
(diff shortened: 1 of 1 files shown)


MicahLyle:
The `@action` decorator was introduced in DRF 3.8.

Djoser would not support 3.7 anymore if this pull request is merged. So maybe when 3.10 comes out and/or support for 3.7 is dropped this could be merged?

https://www.django-rest-framework.org/community/release-notes/#380

dekoza:
Thank you for bringing that up although I'm fully aware of this. This change is part of future version 2.0 - it will remove Python 2 and Django 1 support. (See [ver2](https://github.com/sunscrapers/djoser/tree/ver2) for a taste of things to come) hence the version bump. That being said, I'll close this PR as the work is already applied on other branch.

PR #376, opened 2019-05-09
Fix failing tests for Python 2.7

hawi74:
Current version on `master` does not support Python 2.7.


dekoza:
Thank you, good sir :+1:

PR #377, opened 2019-05-09
Prevent non-staff users from getting other users data

### LINKED ISSUE #371: Get /users/ returns all users by default
Hi djoser team

Awesome app you have developed, been using it for a while and just wanted to double check (and maybe have it clarified in the docs), is the endpoint /users/ supposed to just return all users in the system to anyone? 

Meaning if I have **url(r'^auth/', include('djoser.urls')),** in my urls.py, automatically [my domain]/users/ would return all users in the system to anyone, no auth required by default. Of course if we set PERMISSIONS for user_list it won't show but is it not better to default this permission to either **rest_framework.permissions.IsAdminUser** or **djoser.permissions.CurrentUserOrAdmin** instead of **djoser.permissions.CurrentUserOrAdminOrReadOnly**?

Kind Regards
Baddar

hawi74:
Fixes #371

### DIFF
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -325,6 +325,13 @@ class UserViewSet(UserCreateMixin,
     permission_classes = settings.PERMISSIONS.user
     token_generator = default_token_generator
 
+    def get_queryset(self):
+        qs = super().get_queryset()
+        user = self.request.user
+        if not user.is_staff:
+            qs = qs.filter(pk=user.pk)
+        return qs
+
     def get_permissions(self):
         if self.action == 'create':
             self.permission_classes = settings.PERMISSIONS.user_create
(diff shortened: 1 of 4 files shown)


joshua-s:
@dekoza can this be merged

djoser/views.py
```diff
@@ -325,6 +325,13 @@ class UserViewSet(UserCreateMixin,
     permission_classes = settings.PERMISSIONS.user
     token_generator = default_token_generator
 
+    def get_queryset(self):
+        qs = super().get_queryset()
+        user = self.request.user
+        if not user.is_staff:
```
dekoza: Should be `if not user.is_staff or user.is_superuser:`

PR #382, opened 2019-05-17
Base endpoints docs correction

Chadys:
I fixed some errors in the request params (USERNAME_FIELD was used in some place instead of uid and email field) and I added missing error response code and content


Chadys:
I'm going to make another two pull requests following up on this one : 
- have 'token' field error instead of 'non_field_error' in password-reset-confirm and activation views
- have 'EMAIL_FIELD' instead of 'email' in password-reset and activation-resend views

PR #383, opened 2019-05-17
Remove non_field_error from base endpoints

Chadys:
Following up on #382, this pull request replace 'non_field_error' in password-reset-confirm and activation views with 'token' field error.


dekoza:
This will probably get refactored with `djoser-2.0` but thanks anyway :slightly_smiling_face: Gives me better starting point.

PR #384, opened 2019-05-17
Custom email field for reset and resend views and custom user tests

Chadys:
Following up on #382, this pull request replace 'email' in password-reset and activation-resend views with 'EMAIL_FIELD' field.

I also added some tests that checks djoser compatibility with custom user model by testing different endpoints with custom username, email and required fields.
I needed to use a lot of mock so the test cases aren't pretty, but changing the user model dynamically when the app has already loaded causes some conflicts and I didn't find a better solution. 

If this and #383 are both accepted, I'll handle the merge conflict between those two depending on the first one that get merged.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ This document records all notable changes to djoser.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 
 ---------------------
-`1.6.0`_ (2019-04-02)
+`1.6.0`_ (2019-05-15)
 ---------------------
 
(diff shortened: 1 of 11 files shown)


Chadys:
I can't comment on a deleted file so I'll put it here : I deleted `testproject/__init__.py` because I had the same problem as [here](https://stackoverflow.com/a/51134022), aka *Django model “doesn't declare an explicit app_label”* error caused by empty `__init__.py` in same repository as `manage.py`

dekoza:
Fantastic work, very appreciated. Thank you, @Chadys

PR #387, opened 2019-05-21
added USER_CREATE_PASSWORD_RETYPE option

Chadys:
I think that the user creation endpoint is missing a password confirmation option. Lots of websites ask twice for the password on registration


dekoza:
Good job, thanks :+1:

PR #391, opened 2019-05-25
Minor update to settings docs

Chadys:
I noticed there was one endpoint that was still using the backward compatible url and the setting LOGOUT_ON_PASSWORD_CHANGE doesn't have any effect when you use JWT so I added a note about it.

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -99,7 +99,7 @@ If ``True``, setting new password will logout the user.
 
 .. note::
 
-    Logout will only works with token based authentication.
+    Logout only works with token based authentication.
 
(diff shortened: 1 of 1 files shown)


docs/source/settings.rst
```diff
@@ -97,6 +97,10 @@ If ``True``, setting new password will logout the user.
 
 **Default**: ``False``
 
+.. note::
+
+    Logout will only works with token based authentication.
```
dekoza: `Logout will only work` or `Logout only works` :wink:

Chadys: oopsie, thks

PR #401, opened 2019-06-17
added settings to customize activation mail sending

Chadys:
I think applications should be able to customize the users' activation procedure for their use cases. In the current version, when you want activation  upon registration you also get activation upon email and username change which might not be desired. I added two settings, ``ACTIVATION_NEEDED_ON_EMAIL_UPDATE`` and ``ACTIVATION_NEEDED_ON_USERNAME_UPDATE`` that enable to customize on what event an activation is needed. They are both set to True by default, so as to keep current behaviour. The ``SEND_ACTIVATION_EMAIL`` documentation was also missing the activation on username change.


docs/source/settings.rst
```diff
@@ -38,10 +38,27 @@ SEND_ACTIVATION_EMAIL
 If ``True`` user will be required to click activation link sent in email after:
 
 * creating an account via ``RegistrationView``
-* updating his email via ``UserView``
+* updating ``User.EMAIL_FIELD`` via ``/users/me/`` endpoint (also depends on ``ACTIVATION_NEEDED_ON_EMAIL_UPDATE`` setting)
+* updating ``User.USERNAME_FIELD`` via ``/users/change_username/`` endpoint (also depends on ``ACTIVATION_NEEDED_ON_USERNAME_UPDATE`` setting)
```
Chadys:
```suggestion
* updating ``LOGIN_FIELD`` via ``/users/change_username/`` endpoint (also depends on ``ACTIVATION_NEEDED_ON_USERNAME_UPDATE`` setting)
```
``Set Username`` documentation in ``base_endpoint`` only talks about ``User.USERNAME_FIELD`` but by looking at the code, it seems like it's ``LOGIN_FIELD`` that is edited instead?

Chadys:
Hello, I made this just before version 2.0 came out and it was too late to include my changes in a last 1.X by then. I need to update this PR to become compatible with current djoser version, but I didn't have anymore time to do it then and I still don't have now. Feel free to fork my work and propose your own, up-to-date, PR for that feature. Else I can't guarantee when I'll have enough free time to get back to it.

PR #404, opened 2019-07-15
DRF 3.10 support

ChrisBurch:
Django Rest Framework 3.10.0 was released this morning, and removed the @list_route decorator ([deprecated](https://www.django-rest-framework.org/community/3.8-announcement/#action-decorator-replaces-list_route-and-detail_route) since DRF 3.8). This PR updates those decorators.

It also fixes `test_drf_docs()`, which breaks with DRF 3.10. To do this, I added a default schema class in the example project per the [release notes](https://www.django-rest-framework.org/community/3.10-announcement/#continuing-to-use-coreapi). I'd love a second set of eyes on if this is the correct way to fix this issue (at least during the deprecation window), since I haven't worked with CoreAPI vs OpenAPI schemas in DRF before

- moving from @list_route to @action decorator
- adding a default schema class to the example project
  (https://www.django-rest-framework.org/community/3.10-announcement/#continuing-to-use-coreapi)

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -65,10 +65,10 @@
         },
         "djangorestframework": {
             "hashes": [
-                "sha256:376f4b50340a46c15ae15ddd0c853085f4e66058f97e4dbe7d43ed62f5e60651",
-                "sha256:c12869cfd83c33d579b17b3cb28a2ae7322a53c3ce85580c2a2ebe4e3f56c4fb"
+                "sha256:1f274ebbb930dfa709751fac31b2332b2e662303f8c66d7eacea534702611e34",
+                "sha256:bd389598d052fb79a68904eff35d34303118b5c2abefa65329b5f1a07b571ce1"
             ],
-            "version": "==3.9.4"
+            "version": "==3.10.0"
         },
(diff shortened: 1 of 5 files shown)


ChrisBurch:
Oops... Adding `DEFAULT_SCHEMA_CLASS` isn't backward-compatible with DRF 3.8/3.9 (thanks tox!). I'll see if I can get a better fix for this schema stuff, but I probably won't have more cycles to do that for a few days

dekoza:
Thanks for the effort. We're releasing `djoser-2.0` soon - it will drop support for Py2, Django 1.11 and DRF<=3.7.

ChrisBurch:
Perfect, seems like this PR (once working) will fit right in to those constraints

ozeranskiy:
> Thanks for the effort. We're releasing `djoser-2.0` soon - it will drop support for Py2, Django 1.11 and DRF<=3.7.

what are the tasks for `djoser-2.0`?

dekoza:
Djoser 2.0 has been released, I'm closing this.

ChrisBurch:
🎉 thanks @dekoza

PR #418, opened 2019-08-15
Send the confirmation email after user activation

pkolios:
We recently upgraded to the latest djoser version, and our test suite caught an issue with sending out confirmation emails after a user has been successfully activated. It turns out, the wrong email is sent in djoser's activate view.


dekoza:
Thank you

PR #426, opened 2019-09-04
Tested all the permission schemas

zefciu:
* Changed the settings in testproject to use the `CurrentUserOrAdminOrReadOnly` permissions
* Changed the tests accordingly
* Removed the `user` parameter from client calls that served no purpose

### DIFF
--- a/testproject/testapp/tests/test_user_detail.py
+++ b/testproject/testapp/tests/test_user_detail.py
@@ -3,11 +3,15 @@ from rest_framework import status
 from rest_framework.reverse import reverse
 from rest_framework.test import APITestCase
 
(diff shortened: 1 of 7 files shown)


testproject/testapp/tests/test_user_detail.py
```diff
@@ -27,11 +27,11 @@ def test_user_can_get_own_details(self):
 
         self.assert_status_equal(response, status.HTTP_200_OK)
 
-    def test_user_cannot_get_other_user_detail(self):
```
dekoza: The goal was to NOT allow common users see other users' profiles (at least with default config).

PR #428, opened 2019-09-06
ID_FIELD setting added

hhuseyinpay:
My project requires primary key field of int(id) and guid field for User model. uuid field will used for only public operations. But currently djoser uses only primary key field. This simple changes make possible to configure the public id field.

### DIFF
--- a/djoser/conf.py
+++ b/djoser/conf.py
@@ -27,6 +27,7 @@ class ObjDict(dict):
 
 
 default_settings = {
+    "ID_FIELD": User._meta.pk.name,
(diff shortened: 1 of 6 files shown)


djoser/conf.py
```diff
@@ -27,6 +27,7 @@ def __getattribute__(self, item):
 
 
 default_settings = {
+    "ID_FIELD": User._meta.pk.name,
```
dekoza: I'd change the setting name to `USER_ID_FIELD`

dekoza:
Please also update the docs.

hhuseyinpay:
I think it's done

PR #434, opened 2019-10-25
Use serializer user before request user for password validation

chrislondon-noble:
There will always be a user in the request object.  By default that user is AnonymousUser.  Django has built-in password validation that can valiate from the User object but is moot when validating from the AnonymousUser object.  Switching the order to first use the serializer's user allows for the use of the built-in Django validator django.contrib.auth.password_validation.UserAttributeSimilarityValidator.  Also wrap the self.user in a getattr just in case.


dekoza:
Can you provide a test that supports this change?

PR #436, opened 2019-11-05
Fix base_endpoints.rst User Activate table

alextat120:
Correct syntax error that was preventing the table from rendering properly.


dekoza:
thanks

PR #444, opened 2019-11-21
Fix for CurrentUserOrAdminOrReadOnly permission

mmirandaaraiza:
The CurrentUserOrAdminOrReadOnly permission should inherit from BasePermission.
The has_object_permission method only runs if the has_permission method has passed validation. Currently the permission will always require the user to be logged in, even if the request method is in SAFE_METHODS.


dekoza:
The problem might be a bit deeper than just that. But for now the "CurrentUser" part is what gives the assumption that the user is logged in.

PR #451, opened 2019-12-19
Fix activation email not being send; Fixes #430

### LINKED ISSUE #430: User activation logic does not follow docs
According to [the docs](https://djoser.readthedocs.io/en/latest/settings.html#send-activation-email), a user is "required to click activation link sent in email" after updating their email.

This is not happening and further research into the code reveals this different logic:
https://github.com/sunscrapers/djoser/blob/c5a6f85efb55bc569529c45d98d3992f7ff7fc8b/djoser/views.py#L152

Could you please either patch this or update the docs?

KaczuH:


dekoza:
I feel like this update lacks a test that exposes the bug being patched :wink:

jskitz:
@dekoza this PR should be reverted. I bet whomever submitted the bug had overridden the `UserSerializer` with their own. In that serializer, if the email address has been changed, then `is_active` is set to False, and this is how this email is only sent out when the email changes. By removing this flag, this email is sent on **every** user update, which is not what any one wants here. There is discussion in this PR #557 about this and in this new bug report #546.

PR #454, opened 2019-12-29
Set Password Issue

mrouhi13:
There's an issue on the `set_password` view when these three settings set True, all together.
`PASSWORD_CHANGED_EMAIL_CONFIRMATION`,
`LOGOUT_ON_PASSWORD_CHANGE`,
`CREATE_SESSION_ON_LOGIN`

An exception occurred because we log out the user at the first, so `request.user` set to `AnonymousUser`, next we try to send a confirmation email, what happens? can't find `EMAIL_FIELD` in `AnonymousUser` object.


dekoza:
Nice catch, thank you @mrouhi13 !

PR #457, opened 2020-01-13
German translation

sheepyhollow:
I did aim for consistency within the translated messages.
It should be understandable in any case and I tried to be as polite as the original. Other German users should feel free to adjust for shorter or common lingo if necessary.


dekoza:
Vielen dank, @sheepyhollow :smile: :heart:

PR #484, opened 2020-04-07
use six directly instead of django.utils.six for Django 3.0

snmishra:
Django 3.0 remove django.util.six. The bare six module works fine where it is used. This PR makes that change and also add `six` to Pipfile. Unfortunately, the Pipfile.lock change seems to be rather large.


fmarco:
any update about this one? I'd like to use djoser with django 3.0!

warmwaffles:
Any updates on this?

oclay1st:
@dekoza Could you help us on this, please???

Issue #479, opened 2020-03-24
The field current_password of Set Username endpoint is not required as it should be


truonghn:
Hi Djoser team, Thank you very much for creating such an amazing library. I am having an issue when using Djoser. That is when I used this endpoint: https://djoser.readthedocs.io/en/2.0.3/base_endpoints.html#set-username The `current_password` field is not required as it should be. We could update email without providing the current_password.
if this makes sense, could I submit a PR to add required to the appropriate field?

adriangzz:
I'm having the same issue on Django 3.0.2, djoser 2.0.3 and django 3.8.

krish1237:
I'm having the same issue

krish1237:
Hi, I debugged the source code pertaining the /set_username . It seems to be working fine. Try to reinstall djoser and see if it works. It worked for me that way.

truonghn:
Thank you for your information @krish1237 .
If @krish1237 solution does not work with you, you could try this and see if it works @adriangzz :
So in sub class that inherited from `SetUsernameSerializer`, I added this:
```
class AccountSetUsernameSerializer(SetUsernameSerializer):

    def __init__(self, *args, **kwargs):
        self.fields["current_password"] = serializers.CharField(
            style={"input_type": "password"}
        )
        super().__init__(*args, **kwargs)
```
The reason is in Meta attribute of  `UsernameSerializer` class, it overridden field `fields`, so the field `current_password` was not validated.

krish1237:
```
class CurrentPasswordSerializer(serializers.Serializer):
    default_error_messages = {
        "invalid_password": settings.CONSTANTS.messages.INVALID_PASSWORD_ERROR
    }

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.fields["current_password"] = serializers.CharField(style={"input_type": "password"})

    def validate_current_password(self, value):
        is_password_valid = self.context["request"].user.check_password(value)
        if is_password_valid:
            return value
        else:
            self.fail("invalid_password")
```
This should be the code of CurrentPasswordSerializer for all components extending this class to work as they should. This is because django only considers first extended Serializer as a base class i.e self and any new fields added with ```serializers.Field``` are discarded

truonghn:
That makes perfect sense, @krish1237 . The reason I did not add this to `CurrentPasswordSerializer` class is I don't want to update the code of core lib :) before we confirmed this is the best approach :) . Thanks for your help.

[...]

PR #488, opened 2020-04-20
CurrentPasswordSerializer has been modified to address issue 479

krish1237:
CurrentPasswordSerializer is modified to address the issue 479 of "current password not working as it should". All SerializerClassses extending CurrentPasswordSerializer should work correctly without modifying any code

### DIFF
--- a/djoser/serializers.py
+++ b/djoser/serializers.py
@@ -232,11 +232,13 @@ class PasswordRetypeSerializer(PasswordSerializer):
 
 
 class CurrentPasswordSerializer(serializers.Serializer):
-    current_password = serializers.CharField(style={"input_type": "password"})
-
     default_error_messages = {
         "invalid_password": settings.CONSTANTS.messages.INVALID_PASSWORD_ERROR
     }
+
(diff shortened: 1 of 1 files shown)


dekoza:
@krish1237 please provide a test failing pre-patch and passing post-patch.

adriangzz:
Any update on this?

krish1237:
There are no tests failing pre-patch on the master branch and on post-patch I just modified the code to follow the syntax such that you don't need to include the current_password field like in SetUsernameSerializer and any future classes extending it need not include current_password field.

krish1237:
I followed the syntax in UsernameRetypeSerializer

dekoza:
For this PR to be complete you need to provide a test that would fail without your patch and pass after applying your patch. That's what I meant in my previous comment. You can look for inspiration here: https://github.com/sunscrapers/djoser/tree/master/testproject/testapp/tests

YPCrumble:
To add to this - the issue in #479 is only with a custom user - so adjusting the custom user test should work.

I'd be happy to make a PR @krish1237 if you don't have time.

PR #505, opened 2020-06-13
we can not assume there is always a request, especially DRF OpenAPI j…

jryannel:
…ust tests the view without creating a request currently.


jryannel:
I validated this with my installation. Not sure how to test it best. 

See issue #506

PR #525, opened 2020-09-07
Fix jwt login field docs

tomwojcik:


haxoza (approved):
Good job, let's merge!

PR #527, opened 2020-09-15
Drop support for old Python, Django and DRF versions

dekoza:
Modernized structure of the project - leverage the full potential of `pyproject.toml`:

* use poetry for dependency management and building
* update pre-commit config, add flake8, add some other small hooks
* update isort config - version>=5 does not need to seed third-party-packages list and has a `black` profile
* drop support for outdated Python, Django and DRF versions - hence the version bump (should it be 3.0?)
* tox uses poetry-core to prepare its environments


README.rst
```diff
@@ -39,9 +39,9 @@ Requirements
 
 To be able to run **djoser** you have to meet following requirements:
 
-- Python (3.5, 3.6, 3.7, 3.8)
-- Django (1.11, 2.2, 3.1)
```
tomwojcik, line 43: I don't know if dropping support for Django 1.11 is a good idea. Sounds nice but it doesn't cost us too much to support it and plenty of people are still using it. I know DRF dropped support for 1.11 but their cost is bigger.

.pre-commit-config.yaml
```diff
@@ -1,23 +1,26 @@
...
--   repo: https://github.com/ambv/black
-    rev: stable
-    hooks:
+- repo: https://github.com/psf/black
+  rev: 20.8b1
+  hooks:
     - id: black
-      language_version: python3.7
+
+- repo: https://github.com/pycqa/flake8
```
tomwojcik, line 23: Black is a superset of flake8 rules. Why having both?

haxoza: Not sure if it is superset, e.g. like `max-complexity` setting in Flake8 is more about linting than style formatting.

tomwojcik: Welp, that's right!

README.rst
dekoza: Djoser is a glue layer between Django and DRF. If DRF does not support Django 1.11 (which lifecycle has ended BTW), why should Djoser?

tomwojcik: I didn't know it's officially finished. Then I agree.

testproject/urls.py
```diff
@@ -1,5 +1,5 @@
 try:
-    from django.urls import re_path, include
+    from django.urls import include, re_path
```
tomwojcik, line 2: I added those to support Django 1.x. As we want to drop the support, it should be removed. Here and there's one other place.

tomwojcik (changes requested):

haxoza (changes requested):
Good work! I'm in favour of dropping support for old Django and Python.

However let's move adding poetry to separate pull request before merging this.

testproject/urls.py
dekoza: [These are still supported in Django 3.x](https://docs.djangoproject.com/en/3.1/ref/urls/#re-path)

tomwojcik: Yes, these are still supported, but I added this try/except blog to support Django 1.x.

https://github.com/sunscrapers/djoser/blob/5e68bbbfda03a2d1c3cee414d491a8147a7669af/testproject/urls.py#L1-L5

commit https://github.com/sunscrapers/djoser/commit/a4778357fdc82ddcb1277585676cb8b77d2ab415

dekoza: Done.

tomwojcik (approved):

PR #535, opened 2020-09-24
Travis-ci: added support for ppc64le build along with amd64

sanjaymsh:
Hi,
I have added support for ppc64le build on travis-ci in the branch . The travis-ci build log can be tracked on the link :https://travis-ci.com/github/sanjaymsh/djoser/builds/186542267 . I believe it is ready for the final review and merge.
Please have a look on it and if everything looks fine for you then please approve it for merge.

Thanks !!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,18 @@ dist: xenial
 
 language: python
 
+arch:
+  - amd64
+  - ppc64le
+
 python:
   - "3.5"
   - "3.6"
   - "3.7"
   - "3.8"
 
+before_install:
+  - sudo chown -Rvf $USER:$GROUP ~/.cache/pip/wheels
 install:
   - travis_retry pip install -U tox-travis
 


dekoza:
How exactly is this beneficial for a project that does not require compilation?

sanjaymsh:
Hi @dekoza This package is included in the distros for ppc64le - we are trying to make sure that integration into the distros is pain free by making sure they have a clean bill of health from their development community. This simplifies the integration process for ppc64le based distros.So it may require in future.

Thanks !!

dekoza:
@sanjaymsh but we use Travis only to run the tests, not to build anything. Python packages that don't use specialized C-compiled components are generally distributed as `noarch` packages. I'm sorry but this is a first time I have encountered such request and I want to understand the subject :slightly_smiling_face: 

@haxoza I will need your opinion on this topic.

seth-priya:
@dekoza understand and appreciate you inputs here, thanks!

I think @sanjaymsh 's intention here is to make sure that we are running the full test suites on Power so that we are sure that what is included in the distro has been fully tested and continues to be, not necessarily to re-use the builds from here. It shouldn't take much in terms of investment to enable this for Power as the Travis service is free and I don't expect any maintenance overhead. 

I do agree that if this is all pure python code with no native code or dependencies (and never shall have any) , then it may not be very useful to test it on all architectures, but then we have hit a few cases in the past where we ran into issues originating from failures in the dependencies rather than the main package itself, so may still be good to catch those early.

Thoughts?

dekoza:
@seth-priya thanks for the insight, I really appreciate that.

Right now I can think of one good reason that prevents me from merging this PR. Suppose we add a dependency that will fail only on Power for some reason. This way Travis would report a faulty test pass. Additional configuration would be required to ignore those tests and that would contradict the essence of testing. Plus that would make our team responsible for escalating the problem to original culprit so that our tests could pass again.
On the other hand if the people packaging `djoser` for a Power-based distribution would find the failure during tests **on their side** (specific for the architecture), they could escalate the issue to the original culprit.

In other words: if `djoser` required compilation, I'd gladly add tests for all platforms possible. But since we don't do that, I'm hesitant.

haxoza:
Thank you all for extensive comments.

I totally agree with @dekoza, we should not take the responsibility of maintenance any particular architecture since we are pure Python package.

dekoza:
We won't add it. This is pure-Python package and we don't want to take responsibility for maintaining what should be taken care of either upstream or downstream.

PR #540, opened 2020-09-29
Update serializer to use proper error key

lukzmu:
The selected line crashes when a different value than `non_field_errors` is present in settings.


tomwojcik:
upstream deleted

Issue #541, opened 2020-09-30
permission_denied() got an unexpected keyword argument 'code'


zumoshi:
Error while calling `/users/me/` endpoint with no/wrong auth.

Code in question:
https://github.com/sunscrapers/djoser/blob/35e5a5e75127d7270ab8f7bb13c4f7d0c0ec5a7b/djoser/views.py#L52
Code that calls it (from `django-rest-framework`):
https://github.com/encode/django-rest-framework/blob/be87eb43b348f1f079de81329218b79814b97173/rest_framework/views.py#L333
Change that added the `code` parameter:
https://github.com/encode/django-rest-framework/pull/7306

The only other issue I found about this was self closed: #538

Full trace:
```
Internal Server Error: /auth/users/me/
Traceback (most recent call last):
  File "C:\Users\Bor\AppData\Local\Programs\Python\Python38\lib\site-packages\django\core\handlers\exception.py", line 47, in inner
    response = get_response(request)
  File "C:\Users\Bor\AppData\Local\Programs\Python\Python38\lib\site-packages\django\core\handlers\base.py", line 179, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "C:\Users\Bor\AppData\Local\Programs\Python\Python38\lib\site-packages\django\views\decorators\csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "C:\Users\Bor\AppData\Local\Programs\Python\Python38\lib\site-packages\rest_framework\viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "C:\Users\Bor\AppData\Local\Programs\Python\Python38\lib\site-packages\rest_framework\views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "C:\Users\Bor\AppData\Local\Programs\Python\Python38\lib\site-packages\rest_framework\views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "C:\Users\Bor\AppData\Local\Programs\Python\Python38\lib\site-packages\rest_framework\views.py", line 480, in raise_uncaught_exception
    raise exc
  File "C:\Users\Bor\AppData\Local\Programs\Python\Python38\lib\site-packages\rest_framework\views.py", line 497, in dispatch
    self.initial(request, *args, **kwargs)
[...]

PR #543, opened 2020-10-03
Specified version for djangorestframework

Jeketam:
As noted in #541, djangorestframework version 3.12+ causes TypeErrors on /user/me and /users/ endpoints. 

I noticed that the tests:
test_put_cant_edit_others_attribute, 
test_patch_cant_edit_others_attribute,  
test_put_cant_edit_others_attribute, 
test_unauthenticated_user_cannot_get_user_detail, 
test_user_cannot_get_other_user_detail, 
test_unauthenticated_user_cannot_list_users, 
test_fail_403_without_permission,
test_fail_404_without_permission
also fail with the same TypeError.

Updating the README to specify to use djangorestframework version 3.11.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -87,7 +87,7 @@ requirements manually:
 
 .. code-block:: bash
 
-    $ pip install django djangorestframework
+    $ pip install django djangorestframework==3.11
     $ pip install -r requirements.txt
 
 .. code-block:: bash


dekoza:
Thanks but this part of documentation needs to be rewritten because we switched to `poetry`.

PR #544, opened 2020-10-04
Add poetry support; drop pipenv support; bump version

dekoza:


tomwojcik:
Python3.9 is already released so it might be a good idea to add it to the tox file. Aside from that LGTM.

tomwojcik (dismissed):
Python3.9 is already released so it might be a good idea to add it to the tox file. Aside from that LGTM.

tomwojcik:
I'm not experienced with poetry. I see that we now have redundant information in `pyproject.toml` and `setup.py`. Is it desired?

dekoza:
Yes, `setup.py` can be removed. I overlooked it.

haxoza:
I was a bit afraid that once we merge it then we will receive a bunch of issues (due to lack of `setup.py`) from people who pinned master branch in their requirements. However after many days it seems to be not a problem at all! 🎉

dekoza:
Maybe that's because how setuptools (and pip) work right now - they respect build settings from `pyproject.toml` ([PEP 518](https://www.python.org/dev/peps/pep-0518/)) and run appropriate builder (poetry-core in this case). Anyway, I'm glad too :smile:

Issue #546, opened 2020-10-08
Why send activation email on user update?


akhayyat:
If sending activation emails is enabled (`SEND_ACTIVATION_EMAIL == True`), updating a user currently sends an activation email!

https://github.com/sunscrapers/djoser/blob/master/djoser/views.py#L148:
```python
class UserViewSet(viewsets.ModelViewSet):
    ...
    def perform_update(self, serializer):
        super().perform_update(serializer)
        user = serializer.instance
        # should we send activation email after update?
        if settings.SEND_ACTIVATION_EMAIL:
            context = {"user": user}
            to = [get_user_email(user)]
            settings.EMAIL.activation(self.request, context).send(to)
```

What's the purpose of sending an activation email with an activation link in this scenario (updating a user)? The account is already activated, and will remain activated after the update. A confirmation email would be understandable, but an activation email is puzzling.

dekoza:
How do you see the purpose of activation email? Usually it is used to check if given email is functional. Having users with unusable emails can lead to many problems. Following this logic, if user changes their email, you might want to confirm that it is working properly - hence the activation link. Hope that answers your concern.

akhayyat:
Right. That makes sense when the email is changed. But a user may be updated without changing their email.
In fact, in my case, email is the username. So, changing it requires using the "set username" endpoint. The email field is a read-only field in my user serializer.

But to accommodate the general case, wouldn't it make sense to send an activation email only if the email field is changed?

tartieret:
@dekoza the issue raised by @akhayyat did not exist in previous versions of Djoser, but was modified in this commit:
https://github.com/sunscrapers/djoser/commit/ab3130e08bd65e8bcb4d3b440eb1c00fe8823f4e#diff-ef29656ddb7155dfc495f119a3afdb756b1181b7dee748939b9ec766fdced530
Previously the activation email was only sent IF the user was not active, which kind of makes sense. 

The case when a user changes his/her email is addressed in the UserSerializer here:
https://github.com/sunscrapers/djoser/blob/d12091c4b4ba7b5576e3f53b2b08f6225d55c642/djoser/serializers.py#L24
I don't really see why we would send a confirmation email after each update, and got this issue flagged after it was reported by some of our users following an update of djoser version

haxoza:
Thanks @akhayyat, thanks @tartieret! I agree in general with your comments. I think we should improve this behaviour.

haxoza:
[...]

PR #557, opened 2020-11-08
Update user only sends email confirmation if email field was updated

nicknickel:
I was running into a similar (or same) issue as described here https://github.com/sunscrapers/djoser/issues/546. I had SEND_ACTIVATION_EMAIL=True and was updating a user's information that wasn't the email (i.e. first_name, last_name, etc). This resulted in sending an activation email, which doesn't really make sense since the email isn't being updated.

* Update user only sends email confirmation if email field was updated
  * Updating fields other than the email result in NOT sending an activation email
* Keeps functionality of SEND_ACTIVATION_EMAIL
* All tests passed

### DIFF
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -148,9 +148,8 @@ class UserViewSet(viewsets.ModelViewSet):
     def perform_update(self, serializer):
         super().perform_update(serializer)
         user = serializer.instance
-        email = serializer.validated_data.get("email", None)
         # should we send activation email after update?
-        if settings.SEND_ACTIVATION_EMAIL and email is not None:
+        if settings.SEND_ACTIVATION_EMAIL and not user.is_active:
             context = {"user": user}
             to = [get_user_email(user)]
             settings.EMAIL.activation(self.request, context).send(to)
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -148,8 +148,9 @@ class UserViewSet(viewsets.ModelViewSet):
     def perform_update(self, serializer):
         super().perform_update(serializer)
         user = serializer.instance
+        email = serializer.validated_data.get("email", None)
         # should we send activation email after update?
-        if settings.SEND_ACTIVATION_EMAIL:
+        if settings.SEND_ACTIVATION_EMAIL and email is not None:
             context = {"user": user}
             to = [get_user_email(user)]
             settings.EMAIL.activation(self.request, context).send(to)


akhayyat:
How does this deal with the wording of the activation email? The current wording includes:

```
You're receiving this email because you need to finish activation process on {{ site_name }}.

Please go to the following page to activate account:
...
```

This wording still does not make sense when changing the email of an already active account. Having to use the same email template for both account activation and changing emails still does not make sense to me, hence my suggestion to introduce a new setting for sending email on email changes in #546.

nicknickel:
@akhayyat - I could be interpreting #546 incorrectly. This PR only disables sending activation emails when something other than a user's email address is updated which makes the endpoint match what [the documentation](https://djoser.readthedocs.io/en/latest/settings.html#send-activation-email) indicates. Currently an activation email is sent whenever any attribute of the user is updated, like the first and last name. I can remove the reference to issue 546 if that's not what it is about.

akhayyat:
Right. This does make the behavior consistent with the documentation. So, I guess this is the intended behavior.
Perhaps the default activation template needs to be reworded to accommodate these two cases: creating an account, and updating its email.

jskitz:
What would happen in this case that I pass my email address in the PUT or PATCH, but the email is _exactly_ the same as it was? Wouldn't this trigger the email to still be sent even though it had not really been changed? In version 2.0.3, the code used to look like the following:

https://github.com/sunscrapers/djoser/blob/8f65bfff16577c7fb0f52bbabf5fb69f6809ba62/djoser/views.py#L151

The reason why this works, is that the serializer was setting the `is_active` flag to `False` if the email had been changed as seen here:

https://github.com/sunscrapers/djoser/blob/8f65bfff16577c7fb0f52bbabf5fb69f6809ba62/djoser/serializers.py#L26-L30

So my question, is wouldn't it be better to check if the user `is_active` since the only way you could have done a successful update and gotten to this point would be if the email had changed, which the serializer would then have set `is_active` to `False`?

@akhayyat I really like the idea of having a different email that gets sent out when the email has changed. I am also running into this issue where if I update anything in the user model, it fires off an email. I overloaded the activation email in my project, so my first activation email is also a welcome email, so I can't have this same email being sent for email updates.

@nicknickel are you still working on this? I was thinking I might need to fork and get this fix in there pretty quickly in my own project, but if you are still banging away at it, I might try to get this working with different emails.

Thanks for the work and discussion on this! Much appreciated.

nicknickel:
@jskitz - Thanks! I didn't catch that serializer piece. Basically what you are saying is that, instead of checking for if the email is in the fields that are being updated, just check to see if the user.is_active? I can update my pull request.

jskitz:
@nicknickel yes, that is what I'm saying. I'm guessing that whomever had originally submitted this bug #430 perhaps had overwritten the UserSerializer, or something else was going on with the environment. I have suggested in PR #451 that they simply revert that merge, which would be the fix you have here.

The other item where I think @akhayyat is correct, is that the email that there should be different emails for activation versus just the email address being updated. I want this functionality as well.

I'm not sure who maintains this project (it is not me), but I would approve this PR or revert where the bug actually was introduced. Either way, I'm a bit held up on this one, so will likely be forking and installing from my own repo until this gets merged.

Thanks!

dekoza:
I concur.

maxnowack:
Is there an ETA when this fix will be released?

oscarparedez:
Currently, what is the best solution to this issue? Override partial_update?

Issue #498, opened 2020-06-02
URLs for emails can't be configured for other domains


LyleScott:
It looks like djoser is restricted to using the current site's object for building URLs (ie, `ACTIVATE_URL`) which leaves you with the protocol and domain of the django server. This is a scenario that needs to be configurable, imho, to account for a frontend hosted on some other domain. 

For example, I typically have a Django-based API on the `api` domain (api.foobar.com) and web files served up on the `www` domain (www.foobar.com) via some static-file-only CDN. 

Evidence:

The djoser Activation Email object that produces an evaluated template:
https://github.com/sunscrapers/djoser/blob/84f2ca4e7ce3dd05ce9b145fcc5353e77ebffc05/djoser/email.py#L8

The email template using `protocol`/`domain`:
https://github.com/sunscrapers/djoser/blob/84f2ca4e7ce3dd05ce9b145fcc5353e77ebffc05/djoser/templates/email/activation.html#L22

The django-templated-email handler that fills in the template based off `current_site` (whether you like it or not...):
https://github.com/sunscrapers/django-templated-mail/blob/65e5a34f69d50d4d1b6acdd337f9efcc44b32fae/templated_mail/mail.py#L33

LyleScott:
Would something like this fly? This puts it in the context so it's an _override_:

```diff
diff --git a/djoser/email.py b/djoser/email.py
index 8c01925..0ee0551 100644
--- a/djoser/email.py
+++ b/djoser/email.py
@@ -15,7 +15,15 @@ class ActivationEmail(BaseEmailMessage):
         user = context.get("user")
         context["uid"] = utils.encode_uid(user.pk)
         context["token"] = default_token_generator.make_token(user)
+
+        # Use the [possibly] user supplied values to build links within emails.
+        keys = ("protocol", "domain")
+        for key in keys:
+            settings_value = getattr(settings, key.upper(), None)
+            if settings_value:
+                context[key] = settings_value
         context["url"] = settings.ACTIVATION_URL.format(**context)
+
         return context
 
 
diff --git a/docs/source/base_endpoints.rst b/docs/source/base_endpoints.rst
index ac357e0..a4aaf4e 100644
--- a/docs/source/base_endpoints.rst
+++ b/docs/source/base_endpoints.rst
@@ -38,9 +38,9 @@ User Activate
 
 Use this endpoint to activate user account. This endpoint is not a URL which
 will be directly exposed to your users - you should provide site in your
[...]

PR #566, opened 2020-12-17
fix for issue #498

malikfaiq:


haxoza:
Looks fine though it would be good to have an original behaviour as the default one -- which is guessing those settings from the request object instead of just setting "localhost"-like values.

malikfaiq:
@haxoza I have tried this but, At first, I was not putting it in the default settings but it's causing test cases to fail.

malikfaiq:
Currently, I have created a new pull request and keeping this branch for future use.

Issue #571, opened 2021-01-02
Inactive users returning the same error as wrong credentials


anoited007:
I am using Django 2.2.14 with the configuration below for Djoser 2.1.0 but when trying to get JWT token for an inactive user, it returns the same error as using a wrong password which makes it tricky to differentiate. Am I missing something or that is the expected behaviour? I have look at the error messages defined `constants.py` but the error I expect isn't the one returned. 
I get HTTP STATUS 401 with the detail below
 `{
  "detail": "No active account found with the given credentials"
}`

My configuration is as follows:
```DJOSER = {
    'LOGIN_FIELD': 'email', 
    'SEND_CONFIRMATION_EMAIL': True, 
    'PASSWORD_CHANGED_EMAIL_CONFIRMATION': True, 
    'USER_CREATE_PASSWORD_RETYPE': True, 
    'TOKEN_MODEL': None,  
    'SEND_ACTIVATION_EMAIL': True, 
    "LOGOUT_ON_PASSWORD_CHANGE": False,  
    "PASSWORD_RESET_SHOW_EMAIL_NOT_FOUND": True,  
    "USERNAME_RESET_SHOW_EMAIL_NOT_FOUND": True,  
    'PASSWORD_RESET_CONFIRM_URL': 'account/password/reset/confirm/{uid}/{token}',  
    'USERNAME_RESET_CONFIRM_URL': 'account/username/reset/  /{uid}/{token}', 
    'ACTIVATION_URL': 'account/activate/{uid}/{token}', 
}

I am also using `AUTHENTICATION_BACKENDS = ['django.contrib.auth.backends.AllowAllUsersModelBackend']`

**Comment by jk31** [comment]
Hello @anoited007,

have you found a solution yet? 

I have the same problem. I plan to use the `AllowAllUsersModelBackend` and hoped that djoser supports this workflow. 

While my custom login view allows an inactive user to login, djoser's token login view return 400. 

Moreover, djoser's activation email does not work for inactive users who were logged in with my custom login view (and logged out before activation). Also, requests to base endpoints like `/users/me/` are not possible with inactive users.

**Comment by 3skr0** [comment]
Yes.

I succeed by doing my own view, own serializer. I'm not using the custom settings.

urls.py
```
    path('api/v1/auth/', include('djoser.urls')),
    path('api/v1/auth/jwt/create/token/', core.views.CustomTokenObtainPairView.as_view(),name='token_obtain_pair'),
[...]

PR #578, opened 2021-01-14
Error message if inactive account for email activation requirement

dhakal0kushal:


3skr0:
does it work?

dhakal0kushal:
yeah, it does. there is an error message for inactive users in [serializers.py](https://github.com/sunscrapers/djoser/blob/master/djoser/serializers.py#L107) `inactive_account": settings.CONSTANTS.messages.INACTIVE_ACCOUNT_ERROR` but it's not checking if the user is inactive [here](https://github.com/sunscrapers/djoser/blob/master/djoser/serializers.py#L125)
You can check it in your codebase ^_^

3skr0:
I have tested, but it doesn't looks like to work on my side but i might miss something.

What i did:

`
# in app/serializer.py
class CustomTokenCreateSerializer(TokenCreateSerializer):
    def validate(self, attrs):
        print("xxx")
        password = attrs.get("password")
        params = {djoser.conf.settings.LOGIN_FIELD: attrs.get(djoser.conf.settings.LOGIN_FIELD)}
        self.user = authenticate(
            request=self.context.get("request"), **params, password=password
        )
        if not self.user:
            self.user = User.objects.filter(**params).first()
            if self.user and not self.user.check_password(password):
                self.fail("invalid_credentials")
        if self.user and self.user.is_active:
            return attrs
        if self.user and not self.user.is_active:
            self.fail("inactive_account")
        self.fail("invalid_credentials")

And in settings.py:
DJOSER = {
....
    'SERIALIZERS': {
       'token_create': 'app.serializers.CustomTokenCreateSerializer',
    },
`
but i never see my print(). I guess i'm missing something.

dhakal0kushal:
you'll need to write your custom views.py to use custom serializer and custom urls.py to use that custom views.py

3skr0:
I do not fully understand why.
```
       'token_create': 'app.serializers.CustomTokenCreateSerializer',
```

is not supposed to indicate which serializer to call ?

dhakal0kushal:
are you using JWT for authentication?

3skr0:
Yes. I 've just posted my solution here #571  
[post](https://github.com/sunscrapers/djoser/issues/571#issuecomment-790367556)

PR #582, opened 2021-02-03
Added a settings variable to enable/disable the user delete method

jk31:


jk31:
Today I have realized that you can change the permission settings in the djoser settings to achieve the same results, but somehow my permission does not work.

Here I want only allow staff users to use the delete method for users. For testing, I have created a normal user and try to delete with the same account. With changed permissions this shouln't work. But the _non-staff_ user was able to delete himself.

```python
# main.permissions.py
class OnlyAdminPermission(permissions.BasePermission):
    def has_object_permission(self, request, view, obj):
        user = request.user
        return user.is_staff
```

````python
# settings.py
DJOSER = {
    # "USER_DELETE_DISABLE": True,
    'PASSWORD_RESET_CONFIRM_URL': '#/password/reset/confirm/{uid}/{token}',
    'PERMISSIONS': {
        'user_delete': ['main.permissions.OnlyAdminPermission'],
    }
}
```
First I thought that my permission is not loaded correctly, but I used breakpoints in the djoser source code and saw that my permission is used during the delete method.

Hope someone can help me.

PR #621, opened 2021-07-03
Add option to not send activation email after update

olksndrdevhub:


dekoza:
already taken care of with other PRs

PR #622, opened 2021-07-03
removed duplicate code

aboryone:
Removed duplicate code through base class creation.


tomwojcik:
It's more of a configuration duplication rather than business logic duplication so imo it's fine as is. Thanks.

PR #638, opened 2022-01-01
send activation email only when email is changed

alirezaAsadi2018:
Hi! I am using your perfect library for some time and I really thought there are still problems sending redundant emails when a user wants to update his/her info. I changed the code to only send activation email when user tries to change his/her email, as what is done in serializer and used that in views too. This PR is related to #546

### DIFF
--- a/djoser/serializers.py
+++ b/djoser/serializers.py
@@ -24,10 +24,12 @@ class UserSerializer(serializers.ModelSerializer):
 
     def update(self, instance, validated_data):
         email_field = get_user_email_field_name(User)
+        instance.email_changed = False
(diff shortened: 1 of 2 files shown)


alirezaAsadi2018:
@dekoza could you please take a look at my PR?

Acel-01:
> Hi! I am using your perfect library for some time and I really thought there are still problems sending redundant emails when a user wants to update his/her info. I changed the code to only send activation email when user tries to change his/her email, as what is done in serializer and used that in views too. This PR is related to #546

Thanks. Was able to resolve the same issue on my end with this

Issue #646, opened 2022-01-08
Is there a way to customize HTTP Response Content (More Specifically, Code)? 


ChangyuYan:
For Password Reset, `POST http://localhost:8000/auth/users/reset_password/`, 

1. First of all, the response header doesn't seem to tell me whether user is already registered or not. Of course the password reset link only gets send to the ones who are registered, but is there some other way to automatically tell if user is registered beforehand? 

2. One idea for 1 is to return a customized HTTP response, or at least change the HTTP Code. So far, regardless of whether user is registered, the response code is `204`, the response headers are always the same, and there's no response body. Can we customized this? 

Thanks! 

--------

Edit: 

It looks like, in `djoser/djoser/views.py`, line 236-247: 

```py
@action(["post"], detail=False)
    def reset_password(self, request, *args, **kwargs):
        serializer = self.get_serializer(data=request.data)
        serializer.is_valid(raise_exception=True)
        user = serializer.get_user()

        if user:
            context = {"user": user}
            to = [get_user_email(user)]
            settings.EMAIL.password_reset(self.request, context).send(to)

        return Response(status=status.HTTP_204_NO_CONTENT)
```

The function would return the same response regardless of user exists... is there a way to customize/overwrite this? （Or would it be allowed to add another return statement under the if statement with different status?)

I-am-vishalmaurya:
If else will work right here? If user exits send mail or else return other status

ChangyuYan:
> If else will work right here? If user exits send mail or else return other status

Yeah, exactly. I‘ll try to create a PR?

I-am-vishalmaurya:
Yeah sure

dekoza:
Sorry, but this is a security/privacy antipattern because it allows for checking if there is an account for given email.

PR #647, opened 2022-01-09
added return 200 when user is found while resetting password

ChangyuYan:
Trying to resolve [this issue](https://github.com/sunscrapers/djoser/issues/646)


dekoza:
Sorry, but this is a security/privacy antipattern because it allows for checking if there is an account for given email.

PR #662, opened 2022-04-11
More permissive version constraint for pytz

JaimeLennox:
As pytz is date-versioned, a caret requirement will mean any new release will be incompatible without first updating this project, preventing users from upgrading pytz if they wish so. This removes that constraint.


dcopso:
Is pytz actually used by Djoser? Perhaps it can be removed from the dependency list entirely.

tomwojcik:
Thanks. Not a problem anymore.

PR #663, opened 2022-04-11
Move group dev dependencies to dev block

JaimeLennox:
Using groups is a new feature in the latest preview version of poetry but it's not yet released, and still breaks installations (see https://github.com/python-poetry/poetry/issues/4983). The dev dependencies are compatible with both the stable and preview poetry versions.


tomwojcik:
Thanks, Outdated. Sorry it took so long.

PR #674, opened 2022-05-24
update list of supported versions

asaah18:
update the list of supported versions to be up-to-date with the GitHub repository


tomwojcik:
Actually this is outdated as we don't support Django 1.x. I will remove this section altogether. Thanks for pointing that out,

PR #690, opened 2022-08-28
Allow supplying extra arguments for serializer

Ou7law007:
Example:

Assumption: User model requires extra arguments when a new user is created

With the current implementation, there's no way to supply these extra arguments without overriding AND duplicating most of perform_create's code
But now, it's possible override perform_create as follows:

def perform_create(self, serializer): # overrides djoser's perform_create
    # some code to get/calculate the required extra arguments for creating the user
    my_required_argument_to_create_a_user = 3.14159
    some_more_args = request.something # say, a dict
    return super().perform_create(serializer,my_required_argument_to_create_a_user=my_required_argument_to_create_a_user, **some_more_args)

------
For consistency, I added the same logic to perform_update.

### DIFF
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -132,8 +132,8 @@ class UserViewSet(viewsets.ModelViewSet):
     def get_instance(self):
         return self.request.user
 
-    def perform_create(self, serializer):
-        user = serializer.save()
+    def perform_create(self, serializer, *args, **kwargs):
+        user = serializer.save(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


Ou7law007:
@dekoza 
If you don't mind taking a look at this three-liner. I would like it to be merged so I could use it without modifying djoser's code on 3 different platforms, if there's nothing wrong with it, obviously.

If it cannot be merged, let me know so I can find a different workaround.

Thank you

PR #705, opened 2022-12-09
Update settings - permissions

Maxim-Zima:
It was written 'rest_framework.permissions.CurrentUserOrAdmin'. I suppose it is 'djoser.permissions.CurrentUserOrAdmin', because it is  a special permission and there is no such one in rest_framework. BR
Maxim


DmitrySukharev:
Sure. It was correct in previous versions of docs, but the latest docs version is still incorrect

Issue #703, opened 2022-11-24
Please Publish new version to PyPi


adityaprakashgupta:
The version of Djoser on PyPI is 2.1.0
which was released in 2020.
Please update with new commits.

spockNinja:
For now, I'll be pointing at a commit, which is less than ideal.

We have the technology to automate this process. We can use a GitHub action that only triggers on the creation of a new tag. If a PR for that would help get this done, I'd be happy to help.

There still seems to be plenty of community interest around this library. Does anyone know who the bottleneck is and if we can hand the repo off to someone who wants to make sure it is maintained?

tomwojcik:
I will be the maintainer from now on, but I can work on Djoser only for a few hours a month, so don't expect too much. I will automate as much as possible so releases should be seamless. If you find an issue, please follow with a PR. 

2.2.0a0 has been released. https://pypi.org/project/djoser/2.2.0a0/#description Please let me know if it works. The build and release process was a mess and changed drastically since I released Djoser the last time (3 years ago), so some stuff might not work as expected.

tomwojcik:
No feedback wrt 2.2.0a0 so I assume everything is working as expected. 2.2.0 released https://pypi.org/project/djoser/

PR #721, opened 2023-04-13
Return http 204 instead of 400 if no user

### LINKED ISSUE #656: The resend_activation endpoint discloses if a user with a given email exists
The `resend_activation` endpoint returns a 400 response if the given email does not belong to an (inactive) user.

This endpoint re-uses the password reset serializer (#555) but does not respect the `PASSWORD_RESET_SHOW_EMAIL_NOT_FOUND` setting because of these lines:

https://github.com/sunscrapers/djoser/blob/c62371e3f9a8bbad2eaf55ffd0efad6eb6c02f26/djoser/views.py#L208-L209

All settings related to disclosing email default to `False`:

```
PASSWORD_RESET_SHOW_EMAIL_NOT_FOUND
USERNAME_RESET_SHOW_EMAIL_NOT_FOUND
```

`resend_activation` shouldn't break this default.

P.S.

These settings don't work as advertised by the way, setting one to True has the effect of also toggling the other: 
https://github.com/sunscrapers/djoser/blob/c62371e3f9a8bbad2eaf55ffd0efad6eb6c02f26/djoser/serializers.py#L145-L149

jaredleishman:
Resolves #656 and matches what the [documentation](https://djoser.readthedocs.io/en/latest/settings.html#password-reset-show-email-not-found) says.
@tomwojcik are you the current maintainer? I saw your [comment](https://github.com/sunscrapers/djoser/issues/703#issuecomment-1493325630).


tomwojcik (approved):

tomwojcik:
Yes, thanks for the PR. It will make it into the next release.

tomwojcik:
Could you please rebase with master? I need to see if the CI is green.

jaredleishman:
Should be rebased now @tomwojcik.

djoser/views.py
```diff
@@ -202,7 +202,7 @@ def resend_activation(self, request, *args, **kwargs):
         user = serializer.get_user(is_active=False)
 
         if not settings.SEND_ACTIVATION_EMAIL or not user:
-            return Response(status=status.HTTP_400_BAD_REQUEST)
+            return Response(status=status.HTTP_204_NO_CONTENT)
```
jaap3, line 205: Maybe change this to a "positive check"? I.e.:

```python
if settings.SEND_ACTIVATION_EMAIL and user:
    context = {"user": user}
    to = [get_user_email(user)]
    settings.EMAIL.activation(self.request, context).send(to)
```

That way this method has only one point of return, this seems to be more inline with how the other methods are coded.

jaredleishman: Good idea. Updated.

haxoza:
> Resolves #656 and matches what the [documentation](https://djoser.readthedocs.io/en/latest/settings.html#password-reset-show-email-not-found) says.

Just a heads-up - there is other piece of docs that would need to be updated in case of this change:

https://djoser.readthedocs.io/en/latest/base_endpoints.html#user-resend-activation-e-mail

jaredleishman:
@tomwojcik and @jaap3 I've made some changes to introduce a new setting `RESEND_ACTIVATION_SHOW_EMAIL_NOT_FOUND` so the `/users/resend_activation/` endpoint behaves more like the `/users/reset_password/` endpoint.

djoser/serializers.py
```diff
@@ -145,6 +145,7 @@ def get_user(self, is_active=True):
         if (
             settings.PASSWORD_RESET_SHOW_EMAIL_NOT_FOUND
             or settings.USERNAME_RESET_SHOW_EMAIL_NOT_FOUND
+            or settings.RESEND_ACTIVATION_SHOW_EMAIL_NOT_FOUND
```
jaap3, line 148: As noted in #656, I think this logic is flawed. Setting any one of these settings to `True` has the same effect as setting them all to `True`. These settings imply a granularity that is not actually achieved through code.

It's debatable if these settings should even be there. Account Enumeration is widely regarded as a security flaw, and these flows possibly disclosing the existence of an account are an attack vector.

(note, I'm not a maintainer of Djoser, this is just my opinion, I defer any decision about this to the actual maintainers.)

jaredleishman: Duly noted. I am happy to make changes to address more granularity or remove these settings altogether. Not sure how the maintainers want to move forward though.

tomwojcik: > As noted in #656, I think this logic is flawed. Setting any one of these settings to True has the same effect as setting them all to True. These settings imply a granularity that is not actually achieved through code.

I agree but to me it sounds like a follow up. A small refactor is somewhere on the roadmap, I need to get familiar with the codebase again.

> It's debatable if these settings should even be there. Account Enumeration is widely regarded as a security flaw, and these flows possibly disclosing the existence of an account are an attack vector.

Public tools have different requirements than internal tools. Defaults are safe so I see no harm being done by just having them. 

@jaredleishman would you be willing to just drop this new setting? Your changes with views will mitigate 400 if no user exists and I agree that adding `RESEND_ACTIVATION_SHOW_EMAIL_NOT_FOUND` might not be the best way forward.

tomwojcik: @jaredleishman bump

jaredleishman: Hey @tomwojcik, sorry I didn't see your response to this thread until now. I've removed that new setting.

testproject/testapp/tests/test_resend_activation.py
```diff
@@ -8,6 +8,7 @@
 from testapp.tests.common import create_user, mock
 
 from djoser.compat import get_user_email
+from djoser.conf import settings as default_settings
```
tomwojcik, line 11: CI fails because of this unused import. Please just remove it and then we're good to go.

tomwojcik (changes requested):

djoser/serializers.py
tomwojcik: No worries, thanks for pushing the requested changes. LGTM, just code quality step needs to be fixed.

testproject/testapp/tests/test_resend_activation.py
jaredleishman: Removed. Thanks @tomwojcik

tomwojcik:
Thank you!

tomwojcik (approved):

PR #731, opened 2023-05-13
Rewrite activation email template

snnbotchway:
Rewrote the activation email since the previous one could be improved upon

PR #735, opened 2023-05-20
Revert "Rewrite activation email template"

tomwojcik:
Reverts sunscrapers/djoser#731


tomwojcik:
@snnbotchway While I do agree that this template is better, it just occurred to me that this change would break the translations :( It's not worth the tradeoff.

Issue #716, opened 2023-02-09
Duplicated email address leads to 500 error


zvolsky:
I think 2+ users can share same e-mail address.
However when 2+ users with same address are inactive, djoser will fail so there is no possibility to activate such users.

As you can see, only Django, Rest_framework & Djoser are in the traceback, so I think Djoser should be fixed.

The problem is in the `serializers.py`, `class UserFunctionsMixin`, `def get_user()` where the orm call `User._default_manager.get()` is handled for `User.DoesNotExist` but not for `User.MultipleObjectsReturned`.

I think instead of `.get()` we could use `.filter().first()` here (with appropriate removal of try/except).
This would make the user activating possible, the 1st one first, then the next..

Of course the users identification by email and not by username is not good here.
However I think such solution could give some improvement still.

The other question, which I am not able to answer now, is:
Is it possible to fix it in this way for all scenarios where `UserFunctionsMixin` is used?

```
Internal Server Error: /api/v1/users/resend_activation/
Traceback (most recent call last):
  File "/home/mirek/dj/authtemplate/authtemplate/.venv/lib/python3.10/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/home/mirek/dj/authtemplate/authtemplate/.venv/lib/python3.10/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/mirek/dj/authtemplate/authtemplate/.venv/lib/python3.10/site-packages/sentry_sdk/integrations/django/views.py", line 85, in sentry_wrapped_callback
    return callback(request, *args, **kwargs)
  File "/home/mirek/dj/authtemplate/authtemplate/.venv/lib/python3.10/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/home/mirek/dj/authtemplate/authtemplate/.venv/lib/python3.10/site-packages/rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/mirek/dj/authtemplate/authtemplate/.venv/lib/python3.10/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/home/mirek/dj/authtemplate/authtemplate/.venv/lib/python3.10/site-packages/rest_framework/views.py", line 469, in handle_exception
[...]

PR #747, opened 2023-07-02
Draft: fail if the user exists when creating a new one

tomwojcik:
Closes https://github.com/sunscrapers/djoser/issues/716

todo:
- think twice if it's really needed
- add tests for username and email

### DIFF
--- a/djoser/serializers.py
+++ b/djoser/serializers.py
@@ -79,6 +79,12 @@ class UserCreateSerializer(UserCreateMixin, serializers.ModelSerializer):
                 {"password": serializer_error[api_settings.NON_FIELD_ERRORS_KEY]}
             )
 
+        user_filter_kwargs = {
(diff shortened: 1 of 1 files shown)


insspb:
This PR add exctra SQL query for rare case problem with duplicate fields. basically for very custom models. 
So it is at least better to solve such cases in try except block. As try in python is cheap, in 3.11/3.12 is almost have no overhead (Check pycons 2022/2023). Exception for many cases will not even happen. 
Implementation of .exist() query will affect everybody.

tomwojcik:
> This PR add exctra SQL query for rare case problem with duplicate fields. basically for very custom models. So it is at least better to solve such cases in try except block. As try in python is cheap, in 3.11/3.12 is almost have no overhead (Check pycons 2022/2023). Exception for many cases will not even happen. Implementation of .exist() query will affect everybody.

I agree.

PR #751, opened 2023-07-25
czech (cs) translation, initial

zvolsky:
Just a new language (cs) translation.

It is a little unsure if cs_CZ or cs should be used, but czech language is used in Czech Rep. only (except of really ultra small minorities in Romania, Serbia (or maybe Swiss or USA)).
So cs is I think the proper way to go.
And it is compatible with Django and Django Admin - this is probably the 1st point.

Best regards.
Mirek Zvolský

### DIFF
--- a/djoser/locale/cs/LC_MESSAGES/django.po
+++ b/djoser/locale/cs/LC_MESSAGES/django.po
@@ -1,0 +1,130 @@
+# Copyright (C) Sunscrapers
+# This file is distributed under the same license as the djoser package.
+#
+# Translators:
+# Bertram Bühner <github@sheepyhollow.de>, 2020
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: djoser\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2020-01-13 17:40+0100\n"
+"PO-Revision-Date: \n"
+"Last-Translator: Mirek Zvolský <zvolsky@seznam.cz>\n"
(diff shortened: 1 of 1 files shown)


djoser/locale/cs/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,131 @@
...
+#
+# Translators:
+# Bertram Bühner <github@sheepyhollow.de>, 2020
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: djoser\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2020-01-13 17:40+0100\n"
+"PO-Revision-Date: \n"
```
tomwojcik, line 12: Shouldn't it match `POT-Creation-Date`? Weird that's not been autogenerated with `PO-Revision-Date`.

tomwojcik (changes requested):
Thanks for you contribution! Overall LGTM. I added a test that loads the translations. I think that depending on how you generated the translations, it may or may not be generated with `PO-Revision-Date`. To my understanding, it should be matching `POT-Creation-Date` or newer. In this case, matching, probably. Would you be able to fix it?

```
FAILED testproject/testapp/tests/test_translations.py::test_constants_translations_are_up_to_date - ValueError: time data '' does not match format '%Y-%m-%d %H:%M'
```

tomwojcik:
> It is a little unsure if cs_CZ or cs should be used

`cz` is fine :)

PR #755, opened 2023-08-09
compat.py cleanup and type annotations

insspb:
During digging inside djoser internals I discovered that there is some old code, so I desided to start remove piece by piece and to increase overal code quality. @tomwojcik  what you think?


tomwojcik:
Looks good. I will double check a few things soon, but it makes sense. Thank you

PR #764, opened 2023-09-18
Fixed invalid permission paths for CurrentUserOrAdmin

Ou7law007:
Updated the docs to fix the following error:

```
ImportError: Module "rest_framework.permissions" does not define a "CurrentUserOrAdmin" attribute/class
```


tomwojcik (approved):

tomwojcik:
Good catch, thank you!

PR #789, opened 2024-01-17
actually use USER_ID_FIELD in some places it wasnt used

obayemi:
Make usage of USER_ID_FIELD more consistent
right now using an id that is not strictly the `pk` field make email templates fail because they follow the USER_ID_FIELD, while the endpoints expect strictly a pk without regard to the `USER_ID_FIELD` configured in the settings.

it "may" be a breaking change, but it should be a lot more correct, so I'm not completely sure wether or not you should merge this (at least I needed it for my project, and it would be easier for me to merge this rather than have to maintain a fork)

### DIFF
--- a/djoser/email.py
+++ b/djoser/email.py
@@ -13,7 +13,7 @@ class ActivationEmail(BaseEmailMessage):
         context = super().get_context_data()
 
         user = context.get("user")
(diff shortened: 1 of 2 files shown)


phx000:
Any updates on this PR? Would be nice to have this in master.

Issue #795, opened 2024-02-16
DJOSER is giving tokens to users that don't pass the AUTHENTICATION_BACKENDS.authenticate


hisie:
https://github.com/sunscrapers/djoser/blob/4452009f628b59f02704373d5e7f991f1243397f/djoser/serializers.py#L127

When retrieving a token DJOSER check itself if the password match, or not. If password match, only active or not validation is done.

Django authentication backends propose a set of method to do authentication centralised by backend, while respecting the configuration order (first login accepted first groups and permissions given to the user). 

django.contrib.auth.authenticate is used in AUTHENTICATION_BACKENDShttps://github.com/sunscrapers/djoser/blob/4452009f628b59f02704373d5e7f991f1243397f/djoser/serializers.py#L122. This function call,  in the order declared in AUTHENTICATION_BACKENDS, the authenticate method for each backend.

After checking if the login is correct in all authentication backends, if one user without any permission to get token has a user -password that match, DJOSE give the token.

DJOSER should accept the result of authenticate because developpers can add more condition in the backends to allow user be authenticated.

hisie:
After reading the doc ( https://djoser.readthedocs.io/en/latest/settings.html#login-field) I found this setting an overkill feature. If django user model allows setting the login-field, overriding this with another layer while ignoring the authenticate method can give more problems than solutions.

tomwojcik:
> After reading the doc ( https://djoser.readthedocs.io/en/latest/settings.html#login-field) I found this setting an overkill feature. If django user model allows setting the login-field, overriding this with another layer while ignoring the authenticate method can give more problems than solutions.

It literally works the same as Django, as the default of `LOGIN_FIELD` is `USERNAME_FIELD`. I don't understand what your concern is.

Both the title and the OG msg make it sound as if there was a security issue in Djoser. Djoser does NOT allow users to obtain tokens, if they did not pass the `authenticate` check.

Django's default behavior is to pass the provided credentials through all authentication backends until one succeeds. Once a backend successfully authenticates the user, Django doesn't continue to check other backends.

hisie:
> Djoser does NOT allow users to obtain tokens, if they did not pass the authenticate check.

This is not true: I have tokens sent to users not passing the only valid authenticate in my configuration:

I'm going to show the code. Maybe I'm doing something wrong:
Backend authentication configuration:
```AUTHENTICATION_BACKENDS = ["user_profile.auth_backend.MyCustomBackendAuthentication"]```
[...]

PR #810, opened 2024-04-05
Separate LOGIN_FIELD logic into a auth backend to avoid giving tokens…

hisie:
… to non authorized users.

Corrects #795

### DIFF
--- a/djoser/backends.py
+++ b/djoser/backends.py
@@ -1,0 +1,21 @@
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Permission
+from django.contrib.auth.backends import ModelBackend
+from django.db.models import Exists, OuterRef, Q
+from djoser.conf import settings
+
+User = get_user_model()
+
+
+class LoginFieldBackend(ModelBackend):
+
+    def authenticate(self, request, username=None, password=None, **kwargs):
+        if username is None:
+            username = kwargs.get(settings.LOGIN_FIELD)
+        if username is None or password is None:
+            return
+        user = User.objects.filter(**kwargs).first()
+        if user and not user.check_password(password):
+            return
+        if user and user.is_active:
+            return user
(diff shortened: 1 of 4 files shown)


djoser/backends.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def authenticate(self, request, username=None, password=None, **kwargs):
+        if username is None:
+            username = kwargs.get(settings.LOGIN_FIELD)
+        if username is None or password is None:
+            return
+        user = User.objects.filter(**kwargs).first()
+        if user and not user.check_password(password):
+            return
+        if user and user.is_active:
```
haxoza, line 20: This is good to have so djoser is compatible with Django's default auth model backend. However to be fully covered in the future I would either try to call `super().authenticate()` or at least [`self.user_can_authenticate()`](https://github.com/django/django/blob/5.0.4/django/contrib/auth/backends.py#L51-L56).

haxoza (changes requested):
I like overall approach and idea how to fix that. Thank you for implementing that!

I would only slightly improve the way how `is_active` flag is checked and use `ModelBackend.user_can_authenticate` method for that.

@tomwojcik thoughts?

Also, this change is not fully backend compatible so it would be good to bump at least minor version but may be we should also consider adding that to next major version.

hisie (commented):
Thank you @haxoza. I have leave a comment of why I have chose to not use "standard django methods".

I know this change is incompatible with current djoser, but i have not found a way of recovering standard django auth backend chain check without adding a new django backend.

djoser/backends.py
tomwojcik: I'd definitely release the fix in 2.3 as it's a minor security issue, but at the same time I'm not sure if these particular changes are actually needed.

IMO the changes from https://github.com/sunscrapers/djoser/commit/8f65bfff16577c7fb0f52bbabf5fb69f6809ba62 are incorrect. If the user didn't pass `authenticate`, then they shouldn't be able to login.

If you need a custom behavior, it's up to the `ModelBackend` and user manager implementation (like @hisie did). Handling it by Djoser is a shortcut.

Issue #816, opened 2024-04-15
TokenDestroyView has no serializer class


EvilPug:
Hello! I was generating an openapi schema with drf-spectacular using following command:

```
python ./application/manage.py spectacular --color --file ./application/schema.yml
```

and have encountered an error:
```
.venv/lib/python3.9/site-packages/djoser/views.py:30: Error [TokenDestroyView]: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
```

I guess the solution would be to add empty serializer getters to the TokenDestroyView (if you don't need real ones):

```
def get_serializer(self, *args, **kwargs):
    pass

def get_serializer_class(self):
    pass
```

I was using the latest version of djoser (2.2.2)

haxoza:
Thank you for point that out! I think it makes sense. When I use tooling like that I usually return just empty `Serializer` class from DRF and it works well.

PR for is welcome!

EvilPug:
@haxoza Greetings! Did as you said: added empty ```Serializer``` to views instead of using methods.

tomwojcik:
released in 2.2.3. Thank you!

PR #817, opened 2024-04-19
Added empty serializer classes to fix schema generation (#816)

EvilPug:
- Added empty serializer classes
- Fixed a typo: SingupRequestView -> SignupRequestView

Fixes:
#816

### DIFF
--- a/djoser/webauthn/views.py
+++ b/djoser/webauthn/views.py
@@ -30,7 +30,7 @@ from .utils import create_challenge
 User = get_user_model()
 
 
-class SingupRequestView(APIView):
+class SignupRequestView(APIView):
     permission_classes = (AllowAny,)
     serializer_class = Serializer
 
--- a/djoser/webauthn/views.py
+++ b/djoser/webauthn/views.py
@@ -3,6 +3,7 @@ from django.shortcuts import get_object_or_404
 from rest_framework import status
 from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
+from rest_framework.serializers import Serializer
 from rest_framework.settings import api_settings
 from rest_framework.views import APIView
 from webauthn import (
@@ -31,6 +32,7 @@ User = get_user_model()
 
 class SingupRequestView(APIView):
     permission_classes = (AllowAny,)
+    serializer_class = Serializer
 
     def post(self, request):
         serializer = WebauthnSignupSerializer(data=request.data)
@@ -95,6 +97,7 @@ class SignupView(APIView):
 
 class LoginRequestView(APIView):
     permission_classes = (AllowAny,)
+    serializer_class = Serializer
 
     def post(self, request):
         serializer = WebauthnLoginSerializer(data=request.data)
--- a/djoser/views.py
+++ b/djoser/views.py
@@ -5,6 +5,7 @@ from rest_framework import generics, status, views, viewsets
 from rest_framework.decorators import action
 from rest_framework.exceptions import NotFound
 from rest_framework.response import Response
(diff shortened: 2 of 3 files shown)


djoser/webauthn/views.py
```diff
@@ -29,8 +30,9 @@
 User = get_user_model()
 
 
-class SingupRequestView(APIView):
+class SignupRequestView(APIView):
     permission_classes = (AllowAny,)
+    serializer_class = Serializer
```
haxoza, line 35: Shouldn't it set the serializer class used in the `post()` method, in this case `WebauthnSignupSerializer`?

djoser/webauthn/views.py
```diff
@@ -95,6 +97,7 @@ def post(self, request, ukey):
 
 class LoginRequestView(APIView):
     permission_classes = (AllowAny,)
+    serializer_class = Serializer
```
haxoza, line 100: Same as the above, shouldn't it be `WebauthnLoginSerializer`?

haxoza (commented):
Thank you for submitting this PR!

Overall looks good and it probably works in the use case mentioned in #816.

However I'm not sure if in case of Webauthn views there should be set actual serializer classes used in the code - instead of empty ones.

djoser/webauthn/views.py
EvilPug: Fixed

djoser/webauthn/views.py
EvilPug: Yes, I think this would be correct. Fixed

haxoza (approved):

djoser/webauthn/views.py
```diff
@@ -95,6 +96,7 @@ def post(self, request, ukey):
 
 class LoginRequestView(APIView):
     permission_classes = (AllowAny,)
+    serializer_class = WebauthnLoginSerializer
 
     def post(self, request):
         serializer = WebauthnLoginSerializer(data=request.data)
```
tomwojcik, line 102: please use `self.serializer_class` instead

djoser/webauthn/views.py
```diff
@@ -29,8 +29,9 @@
 User = get_user_model()
 
 
-class SingupRequestView(APIView):
+class SignupRequestView(APIView):
     permission_classes = (AllowAny,)
+    serializer_class = WebauthnSignupSerializer
 
     def post(self, request):
         serializer = WebauthnSignupSerializer(data=request.data)
```
tomwojcik, line 37: please use `self.serializer_class` instead

tomwojcik (changes requested):
`APIView` doesn't really have `serializer_class` attribute, but that's the approach that's already used in other views as well, so I guess it's fine for now.

I've created this discussion. Feel free to share your feedback.

https://github.com/sunscrapers/djoser/discussions/818

Thanks for contributing! Only small nitpicks to address.

djoser/webauthn/views.py
EvilPug: you mean in class scope? but it has no self

tomwojcik: No, I mean

```suggestion
        serializer = self.serializer_class(data=request.data)
```

EvilPug: Oh, got you! Done!

djoser/webauthn/views.py
EvilPug: Done!

tomwojcik (approved):

tomwojcik:
Thank you! Sorry it took so long.

PR #820, opened 2024-05-04
Release/3.0.0

tomwojcik:
todos:
- update docs
- update comments / docstrings, if needed
- move tests to upstream so the diff is smaller and more reliable (unused mocks etc)
- some tests are mediocre at best. Compare the copy-pasted `post` methods body vs upstream ModelViewSet actions
- in the upstream add tests to check the resolved view paths to guarantee they did not change after the migration

### DIFF
--- a/testproject/testapp/tests/test_urls/test_urls.py
+++ b/testproject/testapp/tests/test_urls/test_urls.py
@@ -15,28 +15,23 @@ def test_urls_have_not_changed(settings):
     FILE_PATH = TEST_PATH / "urls_snapshot.json"
     url_patterns = get_resolver().url_patterns
 
-    # Function to normalize URL patterns by removing trailing \Z
-    # otherwise fails in CI
-    def normalize_pattern(pattern):
-        if pattern.endswith("/?$"):
-            pattern = pattern[: -len("/?$")] + r"\Z"
-        return pattern
-
     def get_all_urls(patterns, prefix=""):
         urls = []
         for pattern in patterns:
             if hasattr(pattern, "url_patterns"):
                 urls += get_all_urls(
                     pattern.url_patterns,
-                    prefix + normalize_pattern(pattern.pattern.regex.pattern),
+                    prefix + pattern.pattern.regex.pattern,
                 )
             else:
-                pattern_str = prefix + normalize_pattern(pattern.pattern.regex.pattern)
+                pattern_str = prefix + pattern.pattern.regex.pattern
                 name = pattern.name if pattern.name else None
                 urls.append({"pattern": pattern_str, "name": name})
         return urls
 
     current_urls = sorted(get_all_urls(url_patterns), key=lambda x: x["pattern"])
+    # api-root generates different regex pattern locally vs in CI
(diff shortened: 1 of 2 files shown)


djoser/views/user/user.py
```diff
@@ -0,0 +1,92 @@
...
+
+from djoser import signals, utils
+from djoser.conf import settings
+from djoser.compat import get_user_email
+from django.contrib.auth.tokens import default_token_generator
+
+User = get_user_model()
+
+
+class UserViewSet(viewsets.ModelViewSet):
```
tomwojcik, line 14: should it be split further?

testproject/testapp/tests/test_user_detail.py
```diff
@@ -72,8 +73,11 @@ def test_user_can_get_other_user_detail(self):
 
     def test_user_cant_set_other_user_detail(self):
         login_user(self.client, self.user)
-        response = self.client.get(reverse("user-detail", args=[self.superuser.pk]))
-        self.assert_status_equal(response, status.HTTP_200_OK)
+        response = self.client.patch(
```
tomwojcik, line 76: test body copied from above (dupe), did not match the test name

djoser/urls/authtoken.py
```diff
@@ -1,8 +1,9 @@
...
 
-from djoser import views
+from djoser.views.token.create import TokenCreateView
+from djoser.views.token.destroy import TokenDestroyView
 
 urlpatterns = [
-    re_path(r"^token/login/?$", views.TokenCreateView.as_view(), name="login"),
-    re_path(r"^token/logout/?$", views.TokenDestroyView.as_view(), name="logout"),
+    re_path(r"^token/login/?$", TokenCreateView.as_view(), name="login"),
+    re_path(r"^token/logout/?$", TokenDestroyView.as_view(), name="logout"),
```
haxoza, line 8: We may want to switch to `url()` instead of `re_path` by the way of doing this views split.

djoser/views/user/user.py
haxoza: I think it makes sense to split completely small views.

I think the whole point of splitting is that you can easily get any view, extend it or modify its behavior and inject to your project ONLY what you want.

With this viewset if you don't want to have e.g. update code in you project you can only remove the url pattern, but not the view handler.

Taking a lesson from the community feedback, I would say that viewset is not the best choice for a library.

haxoza (commented):
Overall looks good, however I'm in favor of further splitting!

djoser/views/user/user.py
tomwojcik: I split everything as much as possible, but I doubt `me` can be split further as then we can't register it under the same endpoint, even though it uses a different http method, no?

djoser/urls/authtoken.py
tomwojcik: https://github.com/sunscrapers/djoser/pull/820/commits/44ea180c7d52235a1820d40517a8f90855ea3a80

mdefeche:
Is this release this something that will be done ?

PR #822, opened 2024-05-09
Replace url() with re_path() in the docs

Ash882:
url() is deprecated and has been removed in Django v4


tomwojcik:
You're right, thank you!

PR #861, opened 2025-01-15
Documentation updated for JWT authentication

bilalnaqvi:


zxkeyy:
It should be clarified that setting token_model to none is only required if restframwork.authtoken isn't used otherwise running the authtoken migrations would be the step to take.

tomwojcik:
Thank you!

PR #891, opened 2025-07-14
feat: add swahili locale translation

robinmuhia:


robinmuhia:
@tomwojcik could you kindly review this?

PR #894, opened 2025-07-28
Release/3.0.0

tomwojcik:


djoser/urls/utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Returns:
+        A dispatcher function that can be used directly in URL patterns
+    """
+
+    def dispatcher(request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponseBase:
+        method = request.method
+        if method in method_view_map:
+            view_class = method_view_map[method]
+            return view_class.as_view()(request, *args, **kwargs)
```
haxoza, line 34: do we want to support function based views?

haxoza (approved):
The dispatcher mechanizm looks a bit complicated, but do I get it correctly that it is the cost of splitting the "god" view and keeping the API the same?

Issue #913, opened 2026-02-13
Remove dot in German Translation - thats wrong


janheussner:
**Analysis of the Snippet**

The snippet in question is a segment of a gettext `.po` translation file for the German locale. Specifically, it is this entry:

```po
msgid ""
"You're receiving this email because you need to finish activation process on "
"%(site_name)s."
msgstr ""
"Sie erhalten diese E-Mail, da Sie den Aktivierungsprozess auf %(site_name)s." 
"abschließen müssen."
```

### 1. Code Smells / Antipatterns

- **Line Break in Translated String**: The German translation breaks the sentence across two quoted lines in a way that introduces a period (".") at the end of the first line, which is not present in the English source. This can lead to formatting or grammatical errors in the actual email, as it may result in a sentence fragment or incorrect punctuation.
- **Potential Inconsistency**: The translated string may not match the source in terms of punctuation and sentence structure, which can cause rendering issues or degrade translation quality.

### 2. Recommendations and Improved Example

- **Keep Sentence Structure and Punctuation Consistent**: Ensure the translation has the same sentence structure and punctuation as the original, unless linguistic norms require otherwise.
- **Break Lines Only at Natural Points**: When breaking long strings, do so at logical points (like after commas or clauses) and avoid introducing unintended punctuation.

**Improved Version:**

```po
msgid ""
"You're receiving this email because you need to finish activation process on "
"%(site_name)s."
msgstr ""
"Sie erhalten diese E-Mail, da Sie den Aktivierungsprozess auf %(site_name)s "
"abschließen müssen."
```

PR #915, opened 2026-02-15
Update German translation for activation email

janheussner:
I have fixed the . in german translation for the activation mail.


janheussner:
I have added the fix for https://github.com/sunscrapers/djoser/issues/913

